PR #3, opened 2021-05-31
FEAT: Move PemFileReader to sdk

mariusmihaic:
- Moved **PemFileHandler** from **cli** to sdk **src**
- Moved **IFileHandler** interface from **cli** to sdk **src**
- Deleted **PemFileHandler**::_printFileContent_()
- Made **IFileHandler** pure abstract class so no user can instantiate it
- Small magic number fixes + renamings


cli/commandhandler.h
```diff
@@ -63,11 +65,11 @@ bool init()
 void handleLoadPemFile(const std::map<uint32_t, std::string> &userInputs)
 {
     ih::wrapper::PemHandlerInputWrapper const pemInputWrapper(userInputs);
-    ih::PemInputHandler pemHandler(pemInputWrapper);
+    ih::PemFileReader pemHandler(pemInputWrapper.getPemFilePath());
```
andreibancioiu: The name of the variable can be `pemReader` as well.

cli/commandhandler.h
```diff
@@ -63,11 +65,11 @@ bool init()
...
 
     if (pemHandler.isFileValid())
     {
-        pemHandler.printFileContent();
+        std::cerr << "Bech32 address: " << pemHandler.getAddress().getBech32Address() << "\n";
```
andreibancioiu: Perhaps, on the long run, we'd like to have something similar to a logger with different log levels etc. Can stay as it is for now, of course.

andreibancioiu: About the `reportError` below (sorry for picking it on this PR) - perhaps actually _throwing_ (whether structured or unstructured) _exceptions_ would be more helpful on the long run?

include/account/address.h
```diff
@@ -11,7 +11,7 @@ class Address
 
     Address(bytes const &publicKey);
 
-    Address(std::string bech32Address);
+    Address(std::string const &bech32Address);
```
andreibancioiu, line 14: Here's a [fun discussion](https://stackoverflow.com/questions/10231349/are-the-days-of-passing-const-stdstring-as-a-parameter-over): passing by value (as before) was all right (even preferred IMHO) in your case - your implementation of the constructor now calls the copy constructor of std::string anyway.

include/filehandler/ifilehandler.h
```diff
@@ -10,15 +10,15 @@ class IFileHandler
 public:
     explicit IFileHandler(std::string const &path);
 
-    virtual bool isFileValid() const;
+    virtual bool isFileValid() const = 0;
```
andreibancioiu: Nice! I did not know about [this](https://stackoverflow.com/a/39388910/1475331). Though, there's no _base_ implementation, right (I don't see any)?

include/filehandler/pemhandler.h
```diff
@@ -5,16 +5,15 @@
...
 
 #include "ifilehandler.h"
-#include "wrappers/pem_input_wrapper.h"
 #include "account/address.h"
 #include "internal/internal.h"
 
 namespace ih
 {
-class PemInputHandler : public IFileHandler
+class PemFileReader : public IFileHandler
```
andreibancioiu: 👍 _reader_ makes much more sense. 

Even _IFileHandler_ can be renamed to _IFileReader_, or even _IFile_, I think. `Handler` is a bit imprecise, ambiguous (especially because of the existing term _file handles_) and - if it's sole purpose is to mark a type as `interface` - redundant, since you have the `I` prefix anyway.

src/filehandler/pemhandler.cpp
```diff
@@ -1,44 +1,43 @@
...
 #include <fstream>
 #include <iostream>
 #include <sodium.h>
 
 namespace ih
 {
-PemInputHandler::PemInputHandler(wrapper::PemHandlerInputWrapper const inputWrapper) :
-        IFileHandler(inputWrapper.getPemFilePath()),
-        m_inputData(inputWrapper)
+PemFileReader::PemFileReader(std::string const filePath) :
```
andreibancioiu: What does the namespace / abbreviation _ih_ stand for?

src/filehandler/pemhandler.cpp
```diff
@@ -1,44 +1,43 @@
...
+            IFileHandler::isFileExtension("pem") &&
            (m_fileContent != "");
 }
 
-Address PemInputHandler::getAddress() const
+Address PemFileReader::getAddress() const
 {
     bytes keyBytes = getKeyBytesFromFile();
-    keyBytes.erase(keyBytes.begin(), keyBytes.begin() + 32);
+    keyBytes.erase(keyBytes.begin(), keyBytes.begin() + crypto_sign_PUBLICKEYBYTES);
```
andreibancioiu: _... NUM ..._ or _... LENGTH ..._ - perhaps redefine a constant with a more explicit name, perhaps in a file where we can keep all our constants?

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -345,22 +347,12 @@ TEST(JsonFileHandler, writeOutputFile)
...
-    pemHandler.printFileContent();
-}
-
 TEST(PemFileReader, isPemFileValid_validFile)
 {
     std::map<uint32_t, std::string> inputData;
     inputData[ARGS_TX_IDX_PEM_INPUT_FILE] = "..//testData//keys.pem";
     ih::wrapper::PemHandlerInputWrapper const pemWrapper(inputData);
-    ih::PemInputHandler pemHandler(pemWrapper);
+    ih::PemFileReader pemHandler(pemWrapper.getPemFilePath());
```
andreibancioiu, line 355: 👍  for the tests.

andreibancioiu (approved):

include/filehandler/pemhandler.h
AdoAdoAdo: I would go for IFile, as we will probably add also "write" method(s) to be used when we generate a pem file for example.

cli/commandhandler.h
mariusmihaic: Right, I forgot to refactor the variable names as well, will fix it.

cli/commandhandler.h
mariusmihaic: Good point, I also had in mind to use a logger some time in the future.
About `reportError` you are right, it would be more consistent to just throw errors instead of having mixed: exceptions throw AND error logging. 
For now, we've only refactored SDK for error throwing. We should make it consistent for CLI as well in a future PR.

include/account/address.h
mariusmihaic: My experience with newer C++ standards is pretty limited (I  used to work on embedded C mostly and never used standard libraries) so I used `const &` every time for memory reasons. Recently I've learnt what you mentioned: 

> to pass by value and use std::move(std::string). 

I think in [this](https://github.com/ElrondNetwork/elrond-sdk-erdcpp/pull/4/commits/bed531d55813d0614166a8c186f7388c1a09deaa) PR I took care of it. Will take care of this cases in the future, thx for pointing out !

include/filehandler/ifilehandler.h
mariusmihaic: In this case, there is no base implementation, indeed. However I have some doubts about making it pure virtual, will see in the future if it's going to stay the same.

include/filehandler/pemhandler.h
mariusmihaic: Good point ! Will change its name to _IFile_ !

mariusmihaic: Also renamed pemhandler -> pemreader

src/filehandler/pemhandler.cpp
mariusmihaic: ih = inputhandler. However, I intend to put everything inside SDK APIs into a namespace called `erd`
I left it there because I wanted to take care of namespaces in another PR.

src/filehandler/pemhandler.cpp
mariusmihaic: Good point. I also don't like using those defined values from external library `libsodium`.
In [this](https://elrondnetwork.atlassian.net/browse/EN-10130) ticket I took care of them.

tests/test_cli/test_inputhandler/test_inputhandler.cpp
mariusmihaic: Thank you, appreciate it ! :)
Although, we are missing a lot of unit tests for now.

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #4, opened 2021-06-02
Feat/en 10108/add exceptions

mariusmihaic:
- Added throw exceptions for API SDK  for now
- Fixed some clang warnings
- Refactored some google tests
- Address class can now compute public key from bech32 address and return it


cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
-    if (pemHandler.isFileValid())
+    try
     {
-        std::cerr << "Bech32 address: " << pemHandler.getAddress().getBech32Address() << "\n";
+        ih::wrapper::PemHandlerInputWrapper const pemInputWrapper(userInputs);
+        ih::PemFileReader pemHandler(pemInputWrapper.getPemFilePath());
+        std::cerr << "File loaded successfully! Bech32 address: " << pemHandler.getAddress().getBech32Address() << "\n";
     }
-    else
+    catch (std::exception const &exception)
```
andreibancioiu: These exceptions are caught here - but why not let them reach the caller, instead?

cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
+                     transactionInputWrapper.getData(), transactionInputWrapper.getChainId(),
+                     transactionInputWrapper.getVersion());
+
+            Signer signer(pemFileHandler.getPrivateKey());
+            transaction.applySignature(signer);
+            jsonHandler.writeDataToFile(transaction.getSerializedTransaction());
+        }
+        catch (std::exception const &exception)
+        {
+            std::cerr<<exception.what();
```
andreibancioiu: Again, let them reach the caller, since you only log them anyway, without any specific handling & rethrowing.

cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
-
     ih::JsonFileHandler jsonHandler(transactionInputWrapper.getOutputFile());
-    ih::PemFileReader pemFileHandler(pemInputWrapper.getPemFilePath());
 
-    if (!pemFileHandler.isFileValid())
-    {
-        reportError(ERROR_PEM_INPUT_FILE);
-    }
-    else if (!jsonHandler.isFileValid())
+    if (!jsonHandler.isFileValid())
```
andreibancioiu: Well, perhaps the constructor of `JsonFileHandler` should throw (initialization error), right? Then, we won't need `isFileValid()` checks anymore.

cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
     {
-        Transaction transaction
-                (transactionInputWrapper.getNonce(), transactionInputWrapper.getValue(), transactionInputWrapper.getReceiver(),
-                 pemFileHandler.getAddress(), transactionInputWrapper.getGasPrice(), transactionInputWrapper.getGasLimit(),
-                 transactionInputWrapper.getData(), transactionInputWrapper.getChainId(), transactionInputWrapper.getVersion());
-
-        Signer signer(pemFileHandler.getPrivateKey());
-        transaction.applySignature(signer);
-        jsonHandler.writeDataToFile(transaction.getSerializedTransaction());
+        try
```
andreibancioiu: Don't `try` if there's no `catch` with _specific / particular_ handling.

cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
+        {
+            ih::wrapper::PemHandlerInputWrapper const pemInputWrapper(userInputs);
+            ih::PemFileReader pemFileHandler(pemInputWrapper.getPemFilePath());
+
+            Transaction transaction
```
andreibancioiu: This calls for a `transactionBuilder` (not now). Nice as it is, though :)

cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
-                 pemFileHandler.getAddress(), transactionInputWrapper.getGasPrice(), transactionInputWrapper.getGasLimit(),
-                 transactionInputWrapper.getData(), transactionInputWrapper.getChainId(), transactionInputWrapper.getVersion());
-
-        Signer signer(pemFileHandler.getPrivateKey());
-        transaction.applySignature(signer);
-        jsonHandler.writeDataToFile(transaction.getSerializedTransaction());
+        try
+        {
+            ih::wrapper::PemHandlerInputWrapper const pemInputWrapper(userInputs);
+            ih::PemFileReader pemFileHandler(pemInputWrapper.getPemFilePath());
```
andreibancioiu: `pemReader` (renaming)?

cli/commandhandler.h
```diff
@@ -64,49 +64,54 @@ bool init()
...
+            Transaction transaction
+                    (transactionInputWrapper.getNonce(), transactionInputWrapper.getValue(),
+                     transactionInputWrapper.getReceiver(),pemFileHandler.getAddress(),
+                     transactionInputWrapper.getGasPrice(), transactionInputWrapper.getGasLimit(),
+                     transactionInputWrapper.getData(), transactionInputWrapper.getChainId(),
+                     transactionInputWrapper.getVersion());
+
+            Signer signer(pemFileHandler.getPrivateKey());
+            transaction.applySignature(signer);
+            jsonHandler.writeDataToFile(transaction.getSerializedTransaction());
```
andreibancioiu: `jsonFile` / `outFile`? Also, perhaps, `JsonFile` instead of `JsonFileHandler`? Why `handler`?

cli/inputhandler/arghandler.cpp
```diff
@@ -9,9 +9,9 @@ namespace internal
 // Generic template function to check for user input value.
 // Expects input to be an unsigned long int.
 template<typename T>
-bool isUserInputValid(std::string const arg)
+bool isUserInputValid(std::string const &arg)
 {
-    if ((arg.size() == 0) || (arg.size() > std::numeric_limits<uint64_t>::digits)) return false;
+    if ((arg.empty()) || (arg.size() > std::numeric_limits<uint64_t>::digits)) return false;
```
andreibancioiu, line 14: Perhaps hold a temporary variable called `max_num_digits`?

include/filehandler/pemhandler.h
```diff
@@ -13,22 +13,23 @@ namespace ih
...
     bytes getPrivateKey() const;
 
+protected:
+    bool isFileValid() const override;
+
 private:
     std::string getFileContent() const;
 
-    bytes getKeyBytesFromFile() const;
+    bytes getKeyBytesFromContent(std::string const &content) const;
```
andreibancioiu: Well, since this is a `*Reader`, you can rename the function to `readContent()` or something similar.

include/transaction/transaction.h
```diff
@@ -6,24 +6,24 @@
...
-            std::string chainID,
-            uint64_t version);
-
-    void applySignature(Signer signer);
+            uint64_t const &nonce,
+            std::string const &value,
+            Address const &receiver,
+            Address const &sender,
+            uint64_t const &gasPrice,
+            uint64_t const &gasLimit,
```
andreibancioiu, line 21: Is this intended? 

Usually, light and immutable objects (`Address` objects included) or with _smart moving features implemented by the compiler_ (such as `std::string` - whether it's the `value` up here or the strings held by `Address` objects) can be passed by value (as you did before), for simplicity and also to enable a wide range of optimisations - whether performed by the compiler (unbelievable optimisations nowadays) or as a result of memory caching mechanics - think of looping over one million transactions - operating on pointers pointing to small objects arbitrarily floating in the memory is much worse than operating on copies of those small objects held condensed in a tiny place of the memory. Take this with a grain of salt, of course, I didn't write a benchmark.

andreibancioiu: A bit off-topic - a SDK user might often want to build a large set of transactions (say, hundreds or even thousands) without actually knowing their nonces - that is, the SDK user might want to build the transactions, but delay setting the `nonce` and applying the signatures. I think the same would go for `gasPrice` as well. Perhaps we should handle this scenario in a future PR (currently, the user can only provide these transaction fields within the constructor, and cannot operate on / with partially constructed transactions)?

src/account/address.cpp
```diff
@@ -1,36 +1,56 @@
...
+        throw std::length_error(ERROR_MSG_KEY_BYTES_SIZE);
+}
 
 Address::Address(std::string const &bech32Address) :
         m_pk(),
         m_bech32Address(bech32Address)
-{}
+{
+    if (util::bech32::decode(bech32Address).first != hrp)
+        throw std::invalid_argument(ERROR_MSG_BECH32);
```
andreibancioiu, line 22: 👍 for defining constants for error messages. Perhaps move them to a file called `errors.h`, as well (instead of keeping them in `params.h`)?

src/account/address.cpp
```diff
@@ -1,36 +1,56 @@
...
 
+bytes Address::computePkFromBech32() const
+{
+    bytes pk5BitsPerByte = util::bech32::decode(m_bech32Address).second;
+    unsigned int const pk5BitsSize = pk5BitsPerByte.size();
+
+    auto pkCharPtr = new unsigned char[pk5BitsSize];
+    std::copy(pk5BitsPerByte.begin(), pk5BitsPerByte.end(), pkCharPtr);
+
+    delete[] pkCharPtr;
```
andreibancioiu: I'm a bit out of the loop with this - is it all right to call the `delete` operator, but use `pkCharPtr` afterwards - is the delete deferred to the end of the scope or something like that?

Also, why the copy of `pk5BitsPerByte` to `pkCharPtr`? Might [data()](https://www.cplusplus.com/reference/vector/vector/data/) and [this cast](https://stackoverflow.com/a/4254644/1475331) help perhaps (not sure, in this case)?

src/filehandler/ifilehandler.cpp
```diff
@@ -18,17 +19,17 @@ bool IFileHandler::fileExists() const
...
 }
 
 std::string IFileHandler::getFileExtension() const
 {
-    std::string ext = "";
+    std::string ext;
 
-    if (m_filePath.find_last_of(".") != std::string::npos)
-        ext = m_filePath.substr(m_filePath.find_last_of(".") + 1);
+    if (m_filePath.find_last_of('.') != std::string::npos)
```
andreibancioiu: Sorry for nitpicking - perhaps add a temporary variable called `dot_position` - not for optimisation, of course, but for readability :)

src/filehandler/pemhandler.cpp
```diff
@@ -1,40 +1,50 @@
...
+#include <iomanip>
 
 namespace ih
 {
-PemFileReader::PemFileReader(std::string const filePath) :
+PemFileReader::PemFileReader(std::string const &filePath) :
         IFileHandler(filePath)
 {
-    m_fileContent = getFileContent();
+    if (PemFileReader::isFileValid())
```
andreibancioiu: Why not `throw` when the file is not valid?

src/filehandler/pemhandler.cpp
```diff
@@ -1,40 +1,50 @@
...
-    return IFileHandler::fileExists() &&
-            IFileHandler::isFileExtension("pem") &&
-           (m_fileContent != "");
+    bool const fileExists = IFileHandler::fileExists();
+    bool const fileExtensionValid = IFileHandler::isFileExtension("pem");
+
+    if (!fileExists) throw std::invalid_argument(ERROR_MSG_FILE_DOES_NOT_EXIST);
+    if (!fileExtensionValid) throw std::invalid_argument(ERROR_MSG_FILE_EXTENSION_INVALID);
+
+    return (fileExists && fileExtensionValid) ;
```
andreibancioiu: ... `return true`? Remove return value, actually, since all non-happy cases result in exception? And rename to `checkFile()`? `checkFileExists` + `checkFileExtension`.

src/transaction/signer.cpp
```diff
@@ -1,19 +1,24 @@
...
+    if (secretKey.size() != crypto_sign_SECRETKEYBYTES)
+        throw std::length_error(ERROR_MSG_KEY_BYTES_SIZE);
+}
 
 std::string Signer::getSignature(std::string const &message) const
 {
     //TODO: Put this code in a wrapper for libsodium library
     unsigned long long const msgLength = message.length();
-    unsigned char *msg = new unsigned char[msgLength + 1ULL];
+    auto *msg = new unsigned char[msgLength + 1ULL];
```
andreibancioiu, line 21: Perhaps add a comment explaining why the need to create a new buffer (`msg`) and copy into it? Or is this just for casting purposes? If so, can it be done with a cast?

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -347,42 +347,56 @@ TEST(JsonFileHandler, writeOutputFile)
     jsonHandler.writeDataToFile(transaction.getSerializedTransaction());
 }
 
-TEST(PemFileReader, isPemFileValid_validFile)
+class PemFileReaderConstructorFixture : public ::testing::Test
```
andreibancioiu, line 349: 👍

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -416,13 +430,22 @@ TEST(PemFileReader, getPublicPrivateKeys_expectSameResultFrom_libsodium)
...
+    ih::PemFileReader pemHandler("..//testData//keys.pem");
 
     std::string pemAddress = pemHandler.getAddress().getBech32Address();
     std::string expectedAdr = "erd1sjsk3n2d0krq3pyxxtgf0q7j3t56sgusqaujj4n82l39t9h7jers6gslr4";
 
     EXPECT_EQ(pemAddress, expectedAdr);
 }
+
+//TODO: This test will be moved in another unit test file, when the unit test ticket is started.
+TEST(Address, getPubKey)
```
andreibancioiu, line 441: 👍

cli/commandhandler.h
mariusmihaic: Right, I also had a discussion with @AdoAdoAdo and decided to move the `try-catch` exceptions in `main`

cli/commandhandler.h
mariusmihaic: Done, I've moved them in main

cli/commandhandler.h
mariusmihaic: Forgot about it, fixed it now!

cli/commandhandler.h
mariusmihaic: Yes, it should be consistent, however in this ticket I wanted to only take care of SDK error throwing, but can do it there as well

cli/commandhandler.h
mariusmihaic: Moved` try-catch `from here to main.
However, should we really catch every particular error in CLI? Because every error that could happen is because of a user's wrong input. We should not correct them, only report them, right? The user is the one that should take care of handling its own errors.

cli/commandhandler.h
mariusmihaic: Done, fixed it in this PR

cli/commandhandler.h
mariusmihaic: Good point if different type of transactions are going to be created (e.g.: Create transaction with key file/ledger, etc.). I've left a TODO comment there for future development.

cli/inputhandler/arghandler.cpp
mariusmihaic: `std::numeric_limits<uint64_t>::digits` is already a variable inside a namespace. I don't see much benefits of storing it in another variable, especially because it is only being used once and doesn't really reduce readability. 
I would leave it as it is for now, however if you have a different opinion, I could change it, let me know

include/filehandler/pemhandler.h
mariusmihaic: Regarding method: `getKeyBytesFromContent` I somehow disagree. This method **does not read** anything from any file. It gets as input the content of a file, base64 decodes it, and returns a byte array from it.

However, one might argue about `getFileContent` method's name, since this one actually reads the file content's. The only doubt I have about renaming it is because when you see `readContent`, the name doesn't necessarily points that it returns something, but rather it should have an in-out paramter (e.g.: `readContent(std::string& content)`).

include/transaction/transaction.h
mariusmihaic: Very nice explanation, appreciate the details.
The reason I've made them `const &` is because **CLion IDE Clang-Tidy** complained about the parameters, for instance see this case where I've only deleted const & for receiver's Address:
image

I'm also not sure which approach we should take, however your opinion seems solid. Do you think we should pass them by value instead?

mariusmihaic: Regarding constructing transactions without nonce and gasPrice: you are right, good point, we should take care of it in the future. This topic is already on the table from previous discussions with @AdoAdoAdo, I will also add a TODO here to keep track of it.

src/account/address.cpp
mariusmihaic: Thank you ! :smile: 
That's a good point, I'll move them in a separate file.

src/account/address.cpp
mariusmihaic: > I'm a bit out of the loop with this - is it all right to call the delete operator, but use pkCharPtr afterwards - is the delete deferred to the end of the scope or something like that?

No, it is definitely not ok, that was a terrible mistake I've made :D . I've found the mistake while working on [this ticket](https://elrondnetwork.atlassian.net/browse/EN-10130) and fixed it there. However, I'll fix it here as well.

> Also, why the copy of pk5BitsPerByte to pkCharPtr? Might data() and this cast help perhaps (not sure, in this case)?

I've also seen that cast, but didn't try it, since I've used `std::copy` everywhere else. I've tried it now and it works fine, so I'll stick with your suggestion. However, in this [ticket](https://elrondnetwork.atlassian.net/browse/EN-10130) I will wrap/change `convertBits` function to take as parameter `bytes` instead of `unsigned char*`, so I don't have to always make that cast from outside

src/filehandler/ifilehandler.cpp
mariusmihaic: No prob, good point though !
Added a temporary variable here, as you suggested !

src/filehandler/pemhandler.cpp
mariusmihaic: Because the thrown exception was going to be propagated upwards to the caller of the constructor anyway.
But you are right, it's better to actually to use a try-catch-throw block in the constructor. I've changed it

src/filehandler/pemhandler.cpp
mariusmihaic: Yes, that's right, returning (fileExists && fileExtensionValid) makes no sense, I've missed it :(
I've changed the method name from _isFileValid -> checkFile_.
However, IMHO `fileExists` and `isFileExtension` are better names.If you disagree, we can change it

src/transaction/signer.cpp
mariusmihaic: Yes, the need for the new buffer is for casting purpose only -> since libsodium works with `unsigned char*` instead of `byte array`.
I'll use [this](https://github.com/ElrondNetwork/elrond-sdk-erdcpp/pull/4#discussion_r646188004) suggestion and change the current implementation with a cast in [this](https://elrondnetwork.atlassian.net/browse/EN-10130) PR

cli/commandhandler.h
mariusmihaic: Initially, I used this class for a completely different purpose and it made sense to name it Handler. 
However, the name doesn't suit anymore. I've changed its name to `JsonFile`

cli/inputhandler/jsonfile.cpp
```diff
@@ -0,0 +1,43 @@
...
+        throw;
+    }
+}
+
+bool JsonFile::checkFile() const
+{
+    if (!IFile::fileExists()) throw std::invalid_argument(ERROR_MSG_FILE_DOES_NOT_EXIST);
+    if (!IFile::isFileExtension("json")) throw std::invalid_argument(ERROR_MSG_FILE_EXTENSION_INVALID);
+
+    return true;
```
andreibancioiu: Perhaps we can drop the `bool` from the signature - and perhaps even drop the function `checkFile()` from the `IFile` interface, and only keep it in implementations?

cli/inputhandler/jsonfile.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+void JsonFile::writeDataToFile(std::string const &data)
+{
+    std::ofstream myFile(IFile::getFilePath());
+
+    if (checkFile())
```
andreibancioiu: I see we check the file again - also for its existence. Not sure if we should allow it to be missing, instead, and simply create it (instead of expecting an empty file). Oh, but then there would be changes in the constructor, since the constructor always checks for the existence of the file. Perhaps we can add 2 factory methods (`newInputFile()` and `newOutputFile` - similar to opening a file with a specific `FileMode`, after all) instead of the unique constructor and check for existence in the first one and for non-existence in the second one.

Can stay as it is though for the moment, we can think about this later.

src/filehandler/pemreader.cpp
```diff
@@ -24,17 +25,18 @@ PemFileReader::PemFileReader(std::string const &filePath) :
...
-    bool const fileExists = IFileHandler::fileExists();
-    bool const fileExtensionValid = IFileHandler::isFileExtension("pem");
-
-    if (!fileExists) throw std::invalid_argument(ERROR_MSG_FILE_DOES_NOT_EXIST);
-    if (!fileExtensionValid) throw std::invalid_argument(ERROR_MSG_FILE_EXTENSION_INVALID);
+    if (!IFile::fileExists()) throw std::invalid_argument(ERROR_MSG_FILE_DOES_NOT_EXIST);
+    if (!IFile::isFileExtension("pem")) throw std::invalid_argument(ERROR_MSG_FILE_EXTENSION_INVALID);
 
-    return (fileExists && fileExtensionValid) ;
+    return true;
```
andreibancioiu: Same as above, if you think that's suitable.

andreibancioiu (approved):
👍 🚀 
Nice implementation!

cli/commandhandler.h
```diff
@@ -66,47 +66,31 @@ void handleLoadPemFile(const std::map<uint32_t, std::string> &userInputs)
...
-        reportError(ERROR_PEM_INPUT_FILE);
-    }
+    std::cerr << "File loaded successfully! Bech32 address: " << pemReader.getAddress().getBech32Address() << "\n";
 }
 
 void handleCreateSignedTransactionWithPemFile(const std::map<uint32_t, std::string> &userInputs)
 {
-    ih::wrapper::PemHandlerInputWrapper const pemInputWrapper(userInputs);
     ih::wrapper::TransactionInputWrapper const transactionInputWrapper(userInputs);
+    ih::wrapper::PemHandlerInputWrapper const pemInputWrapper(userInputs);
```
AdoAdoAdo, line 97: can you extract two functions from this one for createTransaction that receives the transactionInputWrapper as parameter and another one signTransaction that receives the created transaction and the private key as parameters?
then call those two functions here.
We will probably need to reuse these two extracted functions for other commands as well.

cli/inputhandler/jsonfile.cpp
AdoAdoAdo: I would also go for the input and output file factory methods.
can stay like this for now

cli/inputhandler/jsonfile.h
```diff
@@ -1,21 +1,22 @@
 #pragma warning(disable:4996)
```
AdoAdoAdo: can we remove the warning disable now?

src/transaction/transaction.cpp
```diff
@@ -3,16 +3,17 @@
...
+        std::string const &value,
+        Address const &receiver,
+        Address const &sender,
+        uint64_t const &gasPrice,
+        uint64_t const &gasLimit,
+        std::string const &data,
+        std::string const &chainID,
+        uint64_t const &version)
 {
     m_jsonTransaction[STR_JSON_NONCE] = nonce;
```
AdoAdoAdo, line 18: can you update instead internal class members of parameters type and just initialize those?
then you can use the json type only inside the getSerializedTransaction method (and probably inside a deserialize method)

src/transaction/transaction.cpp
```diff
@@ -28,7 +29,7 @@ Transaction::Transaction(
     m_jsonTransaction[STR_JSON_VERSION] = version;
 }
 
-void Transaction::applySignature(Signer signer)
+void Transaction::applySignature(Signer const &signer)
 {
     std::string const jsonSerialized = getSerializedTransaction();
     std::string const signature = signer.getSignature(jsonSerialized);
```
AdoAdoAdo, line 35: can we also rename the getSerializedTransaction to serialize()?

src/utils/bits.cpp
```diff
@@ -8,18 +9,17 @@ std::vector<uint8_t> convertBits(unsigned char *data, unsigned int const dataLen
     int acc = 0;
     int bits = 0;
     std::vector<uint8_t> ret;
-    int maxv = (1 << toBits) - 1;
+    int maxV = (1 << toBits) - 1;
```
AdoAdoAdo: can you rename to make this mask name clear?
Also maybe use unsigned types for mask.
We should also make sure we have no overflows (e.g if dataLength > sizeof(int) here )

src/utils/errors.h
```diff
@@ -0,0 +1,53 @@
...
+errorMessage const ERROR_MSG_VALUE = "Invalid value.";
+errorMessage const ERROR_MSG_RECEIVER = "Invalid receiver.";
+errorMessage const ERROR_MSG_GAS_PRICE = "Invalid gas price.";
+errorMessage const ERROR_MSG_GAS_LIMIT = "Invalid gas limit.";
+errorMessage const ERROR_MSG_PEM_INPUT_FILE = "Invalid pem file.";
+errorMessage const ERROR_MSG_JSON_OUT_FILE = "Invalid json file.";
+errorMessage const ERROR_MSG_DATA = "Invalid data.";
+errorMessage const ERROR_MSG_SODIUM_INIT = "Could not initialize sodium library.";
+
+errorMessage const ERROR_MSG_BECH32 = "Invalid bech32 address!";
```
AdoAdoAdo: can we  keep the same format? e.g  some of the error messages end with a dot, others with exclamation mark, etc.

src/utils/errors.h
```diff
@@ -0,0 +1,53 @@
+#ifndef ERD_ERRORS_H
+#define ERD_ERRORS_H
+
+#include <map>
+#include <string>
+
+typedef unsigned long errorCode;
+typedef std::string errorMessage;
+
+errorCode const ERROR_NONE = 0UL;
```
AdoAdoAdo, line 10: can we have instead consecutive values for the error codes?

AdoAdoAdo (changes requested):

cli/inputhandler/jsonfile.h
mariusmihaic: Yes, forgot it there. Removed it now

src/transaction/transaction.cpp
mariusmihaic: Yes, this is already taken care of in [this](https://github.com/ElrondNetwork/elrond-sdk-erdcpp/pull/6) PR

src/transaction/transaction.cpp
mariusmihaic: Yes, this is already taken care of in [this](https://github.com/ElrondNetwork/elrond-sdk-erdcpp/pull/6) PR

src/utils/bits.cpp
mariusmihaic: - Used unsigned int for input parameters as well as for masks
- Changed variable name to maxVal
- About : 

> e.g if dataLength > sizeof(int) here

`dataLength` will be removed in future PR, I don't think it would make sense to change it here

src/utils/errors.h
mariusmihaic: Yes, good point, this should've been changed long time ago. Will change it

src/utils/errors.h
mariusmihaic: Yes, will change it

cli/inputhandler/jsonfile.cpp
mariusmihaic: Removed the bool return and made it void.
Not sure about it, since the whole purpose of making `checkFile` pure virtual is to "force" the classes that derive from IFile to implement this method. This way, when creating a new object that derives from IFile, you  are forced in the constructor to check for file validity and throw errors if needed. We can ofc change this when we implement in/out file factories

cli/inputhandler/jsonfile.cpp
mariusmihaic: Good point. We should take care of it later.

cli/commandhandler.h
mariusmihaic: Done

AdoAdoAdo (approved):

PR #5, opened 2021-06-07
En 10130/add wrapper for libsodium

mariusmihaic:
1. Add wrapper for external library ->libsodium, currently only for the following methods:

- `getSignature`
- `getSecretKey`

2. Defined internal parameters for pk/sk/seed/signature bytes length:
- `#define PUBLIC_KEY_BYTES_LENGTH 32U`
- `#define SECRET_KEY_BYTES_LENGTH 64U`
- `#define SEED_BYTES_LENGTH 32U`
- `#define SIGN_BYTES_LENGTH 64U`

3. Changed **convertBits** function signature to take as input parameter `bytes` instead of (`unsigned char*` + `dataLength`)


src/filehandler/pemreader.cpp
```diff
@@ -22,7 +22,7 @@ PemFileReader::PemFileReader(std::string const &filePath) :
             throw std::invalid_argument(ERROR_MSG_FILE_EMPTY);
 
         m_fileKeyBytes = getKeyBytesFromContent(fileContent);
-        if(m_fileKeyBytes.size() != (crypto_sign_PUBLICKEYBYTES + crypto_sign_SEEDBYTES))
+        if(m_fileKeyBytes.size() != (PUBLIC_KEY_BYTES_LENGTH + SEED_BYTES_LENGTH))
```
andreibancioiu: I know that addition is commutative, but perhaps change the order, to reflect the actual structure of a PEM file? :)

src/filehandler/pemreader.cpp
```diff
@@ -41,29 +41,20 @@ bool PemFileReader::checkFile() const
 
 Address PemFileReader::getAddress() const
 {
-    return Address(bytes(m_fileKeyBytes.begin() + crypto_sign_PUBLICKEYBYTES,m_fileKeyBytes.end()));
+    return Address(bytes(m_fileKeyBytes.begin() + PUBLIC_KEY_BYTES_LENGTH, m_fileKeyBytes.end()));
```
andreibancioiu: Well, please use the other constant. At first, we have the SEED.

Also, you can introduce a temporary variable called `pubkey`, and then create your `Address()` object.

src/filehandler/pemreader.cpp
```diff
@@ -41,29 +41,20 @@ bool PemFileReader::checkFile() const
...
 }
 
 bytes PemFileReader::getSeed() const
 {
-    return bytes(m_fileKeyBytes.begin(),m_fileKeyBytes.begin() + crypto_sign_PUBLICKEYBYTES);
+    return bytes(m_fileKeyBytes.begin(), m_fileKeyBytes.begin() + PUBLIC_KEY_BYTES_LENGTH);
```
andreibancioiu: Please use the other constant, for SEED size.

src/filehandler/pemreader.cpp
```diff
@@ -41,29 +41,20 @@ bool PemFileReader::checkFile() const
...
-    bytes seedBytes = getSeed();
-    bytes pkBytes = getAddress().getPublicKey();
-
-    std::copy(seedBytes.begin(), seedBytes.end(), seed);
-    std::copy(pkBytes.begin(), pkBytes.end(), pk);
-
-    crypto_sign_seed_keypair(pk, sk, seed);
-
-    return bytes(sk, sk + crypto_sign_SECRETKEYBYTES);
+    return wrapper::crypto::getSecretKey(pkBytes,seedBytes);
```
andreibancioiu: seed vs. secret key vs. private key, see comments below (CC: @AdoAdoAdo @ccorcoveanu, can you please have a look?)

Furthermore, why does `crypto::getSecretKey()` need the public key? (I will read below).

src/utils/bits.cpp
```diff
@@ -3,17 +3,15 @@
...
 {
     int acc = 0;
     int bits = 0;
     std::vector<uint8_t> ret;
     int maxV = (1 << toBits) - 1;
     int maxAcc = (1 << (fromBits + toBits - 1)) - 1;
 
-    unsigned int ct = 0;
-    for (unsigned char value = *data; ct<dataLength; value = *++data)
+    for (auto const &value : data)
```
andreibancioiu, line 14: Nice shorthand!

src/utils/bits.h
```diff
@@ -4,9 +4,10 @@
 #include <vector>
 #include <cstdint>
 
+#include "internal/internal.h"
+
 namespace util
 {
-std::vector<uint8_t> convertBits(unsigned char *data, unsigned int dataLength,
-                                 int fromBits, int toBits, bool pad);
+bytes convertBits(bytes const &data, int fromBits, int toBits, bool pad);
```
andreibancioiu: 👍

src/wrappers/cryptosignwrapper.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "strchr.h"
+
+#include <sodium.h>
+
+#if \
+    (PUBLIC_KEY_BYTES_LENGTH != crypto_sign_PUBLICKEYBYTES) || \
+    (SECRET_KEY_BYTES_LENGTH != crypto_sign_SECRETKEYBYTES) || \
+    (SEED_BYTES_LENGTH  != crypto_sign_SEEDBYTES) ||           \
+    (SIGN_BYTES_LENGTH != crypto_sign_BYTES)
+#pragma message "Error. Libsodium library was updated. Update define parameters in the wrapper!"
```
andreibancioiu, line 11: 👍

src/wrappers/cryptosignwrapper.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#else
+
+namespace wrapper
+{
+namespace crypto
+{
+std::string getSignature(bytes const &secretKey, std::string const &message)
```
andreibancioiu, line 19: `sign()` perhaps?

src/filehandler/pemreader.cpp
andreibancioiu: Please have a look at:
https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/walletcore/userKeys.ts#L36
https://github.com/ElrondNetwork/elrond-sdk-erdpy/blob/main/erdpy/wallet/signing.py#L16

There, we only construct a keypair when signing.

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -397,22 +397,16 @@ TEST(PemFileReader, getPublicPrivateKeys_expectSameResultFrom_libsodium)
...
-
-
     bytes seedBytes = pemHandler.getSeed();
     bytes skBytes = pemHandler.getPrivateKey();
     bytes pkBytes = pemHandler.getAddress().getPublicKey();
 
-    std::copy(seedBytes.begin(), seedBytes.end(), pemSeed);
-    std::copy(skBytes.begin(), skBytes.end(), pemSk);
-    std::copy(pkBytes.begin(), pkBytes.end(), pemPk);
+    auto pemPk = reinterpret_cast<unsigned char*> (pkBytes.data());
```
andreibancioiu: Please add the "documentation" signing tests. 

https://github.com/ElrondNetwork/elrond-go/blob/master/examples/construction_test.go
https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/walletcore/users.spec.ts#L120

(with those values, with that data, and expect those signatures)

src/wrappers/cryptosignwrapper.cpp
```diff
@@ -0,0 +1,46 @@
+#include "cryptosignwrapper.h"
```
andreibancioiu, line 1: Perhaps remove / hide `getPrivateKey()`, and rename `getSeed()` to `getSecretKey()`?

CC: @AdoAdoAdo, @ccorcoveanu 

https://github.com/ElrondNetwork/elrond-sdk-erdjs/tree/main/src/walletcore

src/wrappers/cryptosignwrapper.cpp
andreibancioiu: `bytes const &secretKey` - which one is this?

I see this code:

```
Signer signer(pemHandler.getPrivateKey());
transaction.applySignature(signer);
```

then:
```
Signer:: Signer(bytes const &secretKey)

std::string Signer::getSignature(std::string const &message) const
{
    return wrapper::crypto::getSignature(m_sk, message);
}
```

andreibancioiu (changes requested):

src/filehandler/pemreader.cpp
mariusmihaic: Yes, good point, will change it.

src/filehandler/pemreader.cpp
mariusmihaic: Nice suggestion, fixed it

src/filehandler/pemreader.cpp
mariusmihaic: Updated it

src/filehandler/pemreader.cpp
mariusmihaic: > Furthermore, why does crypto::getSecretKey() need the public key? (I will read below).

Because libsodium considers SECRET KEY = SEED + PUBLICKEY (64 bytes)
image

src/utils/bits.cpp
mariusmihaic: Thanks ! :moon:

src/wrappers/cryptosignwrapper.cpp
mariusmihaic: > sign() perhaps?

There is another method that I intend to implement in the future, with this prototype:
`void sign(bytes const &secretKey, std::string const &message, std::string &signedMessage)` which will return the concatenated message + signature of the message. The reason I named it **getSignature** is to highlight that returns the signature of the message, without altering it. However, we can change it if you consider is more appropriate this way

mariusmihaic: Regarding naming of the variable, there is my fault for mixing the names. I will use the same naming as in **libsodium**, which is "**SecretKey**".

src/wrappers/cryptosignwrapper.cpp
mariusmihaic: Made it consistent now. It is called SecretKey everywhere.

src/filehandler/pemreader.cpp
mariusmihaic: Update: We will use SECRET KEY = SEED + PUBLICKEY
When trying to initialize a Signer, it will expect a seed as constructor parameter and will compute the secret key out of it internally.

tests/test_cli/test_inputhandler/test_inputhandler.cpp
mariusmihaic: I've added just one test out of those -> it passed !!! :green_circle: 
That's because in [this](https://github.com/ElrondNetwork/elrond-sdk-erdcpp/pull/6/files) PR, transaction initialization will be changed, so will need to do lot of refactoring. 
Therefore, I will add all remaining tests in a future PR.

src/wrappers/cryptosignwrapper.h
```diff
@@ -0,0 +1,23 @@
+#ifndef ERD_WRAPPER_CRYPTO_SIGN_H
+#define ERD_WRAPPER_CRYPTO_SIGN_H
+
+#include <string>
+
+#include "internal/internal.h"
+
+#define PUBLIC_KEY_BYTES_LENGTH 32U
```
AdoAdoAdo: These defines could be renamed to remove "BYTES"
SIGN_BYTES_LENGTH could become SIGNATURE_LENGTH

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -341,7 +341,7 @@ TEST(JsonFileHandler, writeOutputFile)
                             transactionWrapper.getGasPrice(), transactionWrapper.getGasLimit(),
                             transactionWrapper.getData(), transactionWrapper.getChainId(),
                             transactionWrapper.getVersion());
-    Signer signer(pemHandler.getPrivateKey());
+    Signer signer(pemHandler.getSeed());
```
AdoAdoAdo, line 345: remove commented code line 298

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -393,26 +393,20 @@ TEST_F(PemFileReaderConstructorFixture, invalidFile_notExisting)
...
-    std::copy(skBytes.begin(), skBytes.end(), pemSk);
-    std::copy(pkBytes.begin(), pkBytes.end(), pemPk);
+    auto pemPk = reinterpret_cast<unsigned char*> (pkBytes.data());
+    auto pemSk = reinterpret_cast<unsigned char*> (skBytes.data());
+    auto pemSeed = reinterpret_cast<unsigned char*> (seedBytes.data());
 
+    unsigned char sodiumPk[crypto_sign_PUBLICKEYBYTES];
+    unsigned char sodiumSk[crypto_sign_SECRETKEYBYTES];
 
     crypto_sign_seed_keypair(sodiumPk, sodiumSk, pemSeed);
```
AdoAdoAdo: it looks a bit weird to have test for the used library.
Can we adapt the test so we verify the library wrapper instead?

AdoAdoAdo (changes requested):

src/wrappers/cryptosignwrapper.h
mariusmihaic: Done !

tests/test_cli/test_inputhandler/test_inputhandler.cpp
mariusmihaic: Indeed, it is ugly to have commented code, but is left there for future tests that are coming in the next PR. We will remove it from there

include/transaction/signer.h
```diff
@@ -6,7 +6,7 @@
 class Signer
 {
 public:
-    explicit Signer(bytes const &secretKey);
+    explicit Signer(bytes const &seed);
```
andreibancioiu, line 9: @AdoAdoAdo, @ccorcoveanu, in erdjs, we call `UserSecretKey` the first 32 bytes of the PEM file, and the decrypted payload of a wallet JSON file (called "seed" in tweetnacl / nacl / libsodium).

https://github.com/ElrondNetwork/elrond-sdk/blob/master/erdjs/src/walletcore/userKeys.ts

Is it all right?

src/filehandler/pemreader.cpp
```diff
@@ -39,29 +39,20 @@ void PemFileReader::checkFile() const
...
 }
 
 bytes PemFileReader::getSeed() const
 {
-    return bytes(m_fileKeyBytes.begin(),m_fileKeyBytes.begin() + crypto_sign_PUBLICKEYBYTES);
+    return bytes(m_fileKeyBytes.begin(), m_fileKeyBytes.begin() + SEED_BYTES_LENGTH);
 }
 
-bytes PemFileReader::getPrivateKey() const
+bytes PemFileReader::getSecretKey() const
```
andreibancioiu: @AdoAdoAdo, should we keep both `getSeed()` and `getSecretKey()`, here, in the PEM reader? Or remove the latter - seems not be used, right?

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -448,3 +442,35 @@ TEST(Address, getPubKey)
     EXPECT_EQ(adr.getPublicKey(),pubKeyFromPem);
 }
 
+//TODO: - Change this ugly tests in future ticket, because Transaction class is going to be completely changed.
+// -Move these tests in separate file in the tests ticket
+// -Add more tests for different signatures/transactions
+TEST(Signer, getSignature)
```
andreibancioiu, line 420: Can be renamed to `getSignatureForArbitraryData`, and the other one: `getSignatureForTransaction`.

andreibancioiu (dismissed):
Green, but some questions about naming consistency among our SDKs.

src/filehandler/pemreader.cpp
AdoAdoAdo: I would remove the later if not directly used, so at least we avoid confusion of which is the secret key and which is the seed.

include/transaction/signer.h
AdoAdoAdo: In libsodium secret key is called the concatenation of seed and public key, but the public key is only used to save some computation, so I guess it would be alright either way, since it <contains> the same secret.

src/filehandler/pemreader.cpp
mariusmihaic: Agree, I removed it, as well as other unused functions and one test

tests/test_cli/test_inputhandler/test_inputhandler.cpp
mariusmihaic: This will definitely be renamed in the future PR with tests.

tests/test_cli/test_inputhandler/test_inputhandler.cpp
mariusmihaic: After all, I deleted the whole test :D

tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -393,22 +393,6 @@ TEST_F(PemFileReaderConstructorFixture, invalidFile_notExisting)
...
-    bytes pkPem = pemHandler.getAddress().getPublicKey();
-
-    bytes const skWrapper = wrapper::crypto::getSecretKey(seedBytes);
-    bytes const pkWrapper = wrapper::crypto::getPublicKey(seedBytes);
-
-    EXPECT_EQ(pkWrapper, pkPem);
-    EXPECT_EQ(skWrapper, skPem);
-}
-
 TEST(PemFileReader, getSegwitAddress)
```
AdoAdoAdo, line 396: getBech32Address

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #6, opened 2021-06-09
En 10147/add wrapper for json

mariusmihaic:
- Updated external library `json.hpp`. No **need to review it** !!!
- Added wrapper for external **nlohmann** `json.hpp` library 
- Updated Transaction Input Wrapper for CLI
- Updated parameters file
- Added just few (ugly) tests for transaction. More will follow in future ticket.

**Ordered json wrapper** when asked to set a `<key,value>` pair :
- base64 encodes `value` if it is a byte array
- sets `<key,value>` pair otherwise

**Transaction class**:
- Has a default constructor, which sets all fields to default values
- Has another constructor which expects all fields. For "omit empty" fields, user can simply pass `nullptr`
- Can sign/serialize/deserialize transactions, throws errors if not able


tests/test_cli/test_inputhandler/test_inputhandler.cpp
```diff
@@ -338,13 +338,19 @@ TEST(JsonFileHandler, writeOutputFile)
...
+                            transactionWrapper.getSignature(), transactionWrapper.getChainId(),
+                            transactionWrapper.getVersion(), transactionWrapper.getOptions());
+
     Signer signer(pemHandler.getSeed());
-    transaction.applySignature(signer);
-    jsonFile.writeDataToFile(transaction.getSerializedTransaction());
+    transaction.sign(signer);
+    jsonFile.writeDataToFile(transaction.serialize());
+    transaction.deserialize("{\"nonce\":5,\"value\":\"10000000000000000000\",\"receiver\":\"erd10536tc3s886yqxtln74u6mztuwl5gy9k9gp8fttxda0klgxg979srtg5wt\",\"sender\":\"erd1sjsk3n2d0krq3pyxxtgf0q7j3t56sgusqaujj4n82l39t9h7jers6gslr4\",\"gasPrice\":1000000000,\"gasLimit\":50000,\"data\":\"test\",\"signature\":\"62af8fa927e4f1ebd64fb8d7cca8aac9d5d33fefa4b185d44bb16ecefc2a7214304b4654406fe76fa36207fbb91f245586f66500cc554a3eb798faab8c435706\",\"chainID\":\"T\",\"version\":1}");
+    jsonFile.writeDataToFile(transaction.serialize());
```
AdoAdoAdo, line 352: The verification is missing

AdoAdoAdo: The verification is missing

mariusmihaic: Changed the test.
All tests are going to be changed in future PR anyway.

include/account/address.h
```diff
@@ -17,6 +17,8 @@ class Address
 
     explicit Address(std::string const &bech32Address);
 
+    Address& operator = (const Address &address);
```
andreibancioiu: Shouldn't an `Address` object be immutable? Why copy another `Address` into it? Useful for initialization? Then why not use a copy constructor? (sorry if I misunderstood).

Perhaps in the future we can override the comparison operator `==`, instead? (quite useful for SDK clients IMHO).

src/account/address.cpp
```diff
@@ -22,6 +22,14 @@ Address::Address(std::string const &bech32Address) :
         throw std::invalid_argument(ERROR_MSG_BECH32);
 }
 
+Address& Address::operator = (const Address &address)
+{
+    this->m_pk = address.m_pk;
```
andreibancioiu: This should call the copy constructors for `string` and `vector<u8>` under the hood, right? (therefore, no data ownership issues).

Also, as above - why overriding the copy "into" / assignment operator, and why not the copy constructor instead?

andreibancioiu (dismissed):

include/account/address.h
mariusmihaic: That's actually a very good comment. I think I needed it before and left it there.
I think you are right, address objects should be immutable, I will implement comparison operator instead

src/account/address.cpp
mariusmihaic: From : https://www.cplusplus.com/reference/vector/vector/operator=/:
> The copy assignment (1) copies all the elements from x into the container (with x preserving its contents).

However, you are right, address objects should be immutable, I will delete the assignment operator and override the comparison operator

src/account/address.cpp
```diff
@@ -7,39 +7,36 @@
...
         m_bech32Address(bech32Address)
 {
     if (util::bech32::decode(bech32Address).first != hrp)
         throw std::invalid_argument(ERROR_MSG_BECH32);
+
+    m_pk = computePkFromBech32();
 }
 
-Address& Address::operator = (const Address &address)
+bool Address::operator==(const Address &address) const
```
andreibancioiu, line 27: In a future PR, please add tests for this as well.

andreibancioiu (approved):

src/account/address.cpp
mariusmihaic: Yes, I've just added tests for it in [this](https://github.com/ElrondNetwork/elrond-sdk-erdcpp/pull/7) PR

AdoAdoAdo (approved):

PR #8, opened 2021-06-21
En 10080/https client wrapper

mariusmihaic:
Attention! **Do NOT** review external HTTP client library: `external/http/httplib.h`

- Added wrapper for external HTTP client library for **GET** and **POST**
-  Added unit tests
- Each **GET** and **POST** request will return a `Result` structure, containing:
1. Status code (e.g.: 200 = OK)
2. Status message associated with status code 
3. Error boolean 
4. Body string

### DIFF
--- a/src/wrappers/httpwrapper.h
+++ b/src/wrappers/httpwrapper.h
@@ -5,7 +5,6 @@
 
 #define STATUS_CODE_DEFAULT -1
 #define STATUS_CODE_OK 200
-#define STATUS_MSG_OK "Ok"
 
 namespace wrapper
 {
@@ -37,11 +36,7 @@ public:
         if (res)
         {
             status = res->status;
-
-            if (status == STATUS_CODE_OK)
-            {
-                body = res->body;
-            }
+            body = res->body;
         }
         else
         {
@@ -82,7 +77,7 @@ private:
 
     std::string getStatusMessage(int const &status) const
     {
-        return (status == STATUS_CODE_OK) ? STATUS_MSG_OK : std::string(httplib::detail::status_message(status));
+        return std::string(httplib::detail::status_message(status));
     }
 
     httplib::Client m_client;
--- a/src/wrappers/httpwrapper.h
+++ b/src/wrappers/httpwrapper.h
@@ -3,9 +3,8 @@
 
 #include "http/httplib.h"
 
-#define STATUS_CODE_OK 200
-#define STATUS_CODE_BAD_REQUEST 400
 #define STATUS_CODE_DEFAULT -1
+#define STATUS_CODE_OK 200
 #define STATUS_MSG_OK "Ok"
 
 namespace wrapper
@@ -58,7 +57,8 @@ public:
         bool err = false;
         std::string body;
 
-        auto const res = m_client.Post(path.c_str(), message, "text/plain");
+        char const *contentType = "text/plain";
+        auto const res = m_client.Post(path.c_str(), message, contentType);
 
         if (res)
         {
--- a/src/wrappers/httpwrapper.h
+++ b/src/wrappers/httpwrapper.h
@@ -27,34 +27,27 @@ public:
 
     Result get(std::string const &path)
     {
-        int status = STATUS_CODE_DEFAULT;
-        bool err = false;
-        std::string body;
-
         auto const res = m_client.Get(path.c_str());
 
-        if (res)
-        {
-            status = res->status;
-            body = res->body;
-        }
-        else
-        {
-            err = error(res);
-        }
-
-        return Result{status, err, body, getStatusMessage(status)};
+        return wrappedResult(res);
     }
 
     Result post(std::string const &path, std::string const &message)
+    {
+        char const *contentType = "text/plain";
+        auto const res = m_client.Post(path.c_str(), message, contentType);
+
+        return wrappedResult(res);
+    }
+
+private:
+
+    Result wrappedResult(httplib::Result const &res)
     {
         int status = STATUS_CODE_DEFAULT;
         bool err = false;
         std::string body;
 
-        char const *contentType = "text/plain";
-        auto const res = m_client.Post(path.c_str(), message, contentType);
-
         if (res)
         {
             status = res->status;
@@ -67,8 +60,6 @@ public:
 
         return Result{status, err, body, getStatusMessage(status)};
     }
-
-private:
 
     bool error(httplib::Result const &res) const
     {
--- a/src/wrappers/httpwrapper.h
+++ b/src/wrappers/httpwrapper.h
@@ -4,7 +4,13 @@
 #include <string>
 #include "http/httplib.h"
 
+#define STATUS_CODE_OK 200
+#define STATUS_CODE_BAD_REQUEST 400
+#define STATUS_CODE_DEFAULT -1
+#define STATUS_MSG_OK "Ok"
 
+namespace wrapper
+{
 namespace http
 {
 
@@ -13,18 +19,20 @@ struct Result
     int status;
     bool error;
     std::string body;
+    std::string statusMessage;
 };
 
 class Client
 {
 public:
-    explicit Client(std::string const &url): m_client(url.c_str()){}
+    explicit Client(std::string const &url) : m_client(url.c_str())
+    {}
 
     Result get(std::string const &path)
     {
-        std::string body;
-        int status = -1;
+        int status = STATUS_CODE_DEFAULT;
         bool error = false;
+        std::string body;
 
         m_client.Get(path.c_str());
 
@@ -32,7 +40,7 @@ public:
         {
             status = res->status;
 
-            if (res->status == 200)
+            if (status == STATUS_CODE_OK)
             {
                 body = res->body;
             }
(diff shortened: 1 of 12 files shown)


src/wrappers/httpwrapper.h
```diff
@@ -0,0 +1,93 @@
...
+private:
+
+    bool error(httplib::Result const &res) const
+    {
+        return res.error() != httplib::Error::Success;
+    }
+
+    std::string getStatusMessage(int const &status) const
+    {
+        return (status == STATUS_CODE_OK) ? STATUS_MSG_OK : std::string(httplib::detail::status_message(status));
```
AdoAdoAdo: The status messages are pretty standard, don’t think we need to wrap them, just return.

tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -149,3 +150,50 @@ TEST(CryptoWrapper, getSignature)
     EXPECT_EQ(util::stringToHex(signature), expectedSignature);
 
 }
+
+TEST(ClientWrapper, get_validClient)
+{
+    wrapper::http::Client client("https://api.elrond.com");
```
andreibancioiu: Perhaps use `https://gateway.elrond.com` instead? Or even better, `https://testnet-gateway.elrond.com` - so that we don't hit the Mainnet API when running the tests?

tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -149,3 +150,50 @@ TEST(CryptoWrapper, getSignature)
...
+    wrapper::http::Result res =  client.get("/address/erd1l453hd0gt5gzdp7czpuall8ggt2dcv5zwmfdf3sd3lguxseux2fsmsgldz");
```
andreibancioiu: Maybe use a well-known address (e.g. the staking system smart contract), de-serialize the response into a dummy structure (or dictionary) - yes, the test would be more complex, but more insightful as well - and assert for non-zero balance?

tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -149,3 +150,50 @@ TEST(CryptoWrapper, getSignature)
...
+    EXPECT_EQ(res.status, STATUS_CODE_OK);
+    EXPECT_FALSE(res.body.empty());
+    EXPECT_TRUE(res.body.find("success"));
+    EXPECT_EQ(res.statusMessage, STATUS_MSG_OK);
+}
+
+TEST(ClientWrapper, get_invalidClient)
+{
+    wrapper::http::Client client("https://api.elronddd.com");
+    wrapper::http::Result res =  client.get("/address/erd1l453hd0gt5gzdp7czpuall8ggt2dcv5zwmfdf3sd3lguxseux2fsmsgldz");
```
andreibancioiu: Perhaps a more real test-case would be to request a badly formatted, non-bech32 address?

tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -149,3 +150,50 @@ TEST(CryptoWrapper, getSignature)
...
+
+    EXPECT_TRUE(res.error);
+    EXPECT_NE(res.status, STATUS_CODE_OK);
+    EXPECT_TRUE(res.body.empty());
+    EXPECT_NE(res.statusMessage, STATUS_MSG_OK);
+}
+
+TEST(ClientWrapper, post_validClient)
+{
+    wrapper::http::Client client("https://api.elrond.com");
```
andreibancioiu: As above, `https://testnet-gateway.elrond.com`?

tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -149,3 +150,50 @@ TEST(CryptoWrapper, getSignature)
...
+    wrapper::http::Result res =  client.post("/transaction/send", "test");
+
+    EXPECT_FALSE(res.error);
+    EXPECT_EQ(res.status, 400); // 400 = BAD REQUEST
+    EXPECT_FALSE(res.body.empty());
+    EXPECT_TRUE(res.body.find("error"));
+    EXPECT_NE(res.statusMessage, STATUS_MSG_OK);
```
andreibancioiu: Additionally, increase the specificity of the test, with an extra, "positive" assertion? (e,g. the status message equals or contains ...).

tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -149,3 +150,50 @@ TEST(CryptoWrapper, getSignature)
...
+}
+
+TEST(ClientWrapper, post_invalidClient)
+{
+    wrapper::http::Client client("https://api.elronddd.com");
```
andreibancioiu: If you really need this kind of tests, please use a GOOD domain & a BAD subdomain instead (since the domain `elronddd` is not owned by us, and thus the test can be influenced by external actors).

src/wrappers/httpwrapper.h
```diff
@@ -0,0 +1,93 @@
...
+#define STATUS_CODE_DEFAULT -1
+#define STATUS_CODE_OK 200
+#define STATUS_MSG_OK "Ok"
+
+namespace wrapper
+{
+namespace http
+{
+
+struct Result
```
andreibancioiu, line 14: 👍

src/wrappers/httpwrapper.h
```diff
@@ -0,0 +1,93 @@
...
+        return Result{status, err, body, getStatusMessage(status)};
+    }
+
+    Result post(std::string const &path, std::string const &message)
+    {
+        int status = STATUS_CODE_DEFAULT;
+        bool err = false;
+        std::string body;
+
+        char const *contentType = "text/plain";
```
andreibancioiu: Not sure if we want this hardcoded - our APIs might accept this `contentType`, but they will always want `application/json` instead.

Do you think we can create an extra wrapper over this wrapper to offer JSON-friendly (our APIs-friendly), generic (with templates) GET & POST, as well? That is, this extra wrapper would also perform some serialization / de-serialization (if the response is successful) or throw exceptions (otherwise)? Or perhaps that is already on the todo-list :)

E.g.
https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/proxyProvider.ts
https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/apiProvider.ts

andreibancioiu (changes requested):

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: Changed it to `https://testnet-gateway.elrond.com`. I was not aware of them, I followed the documentation from here: https://docs.elrond.com/sdk-and-tools/rest-api/rest-api/

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: I changed the address to : `erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqplllst77y4l`

>de-serialize the response into a dummy structure

I wanted to leave these tests for the ProxyProvider that I'm going to implement in a future ticket. For this particular tests, I just wanted to check that the client wrapper returns a **Result** structure with some valid data.
I can, however, deserialize it and expect more precise/insightful results, it would be better.

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: Yes, that's right, I've adapted the tests to request a badly formatted address

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: Yes, changed it everywhere

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: Yes, that's a good point, I've refactored the tests to check for valid domain and bad subdomains instead

src/wrappers/httpwrapper.h
mariusmihaic: Thanks ! :)

src/wrappers/httpwrapper.h
mariusmihaic: > Not sure if we want this hardcoded - our APIs might accept this contentType, but they will always want application/json instead.

Indeed, this going to be changed in a future PR and  **POST** method will accept content type as an extra parameter.

> Do you think we can create an extra wrapper over this wrapper to offer JSON-friendly (our APIs-friendly), generic (with templates) GET & POST, as well? That is, this extra wrapper would also perform some serialization / de-serialization (if the response is successful) or throw exceptions (otherwise)? Or perhaps that is already on the todo-list :)

Very good point, while working on the next PR I've noticed this and started working on an extra wrapper for our APIs responses.
However, I will not add it here, since in this PR we just want a wrapper dedicated to a client external library. 
Another wrapper will be created in future PR with the Provider, so it's already on the todo-list.

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: Yes, changed it to expect the actual status message

src/wrappers/httpwrapper.h
mariusmihaic: Yes, I've deleted **STATUS_MSG_OK**.

AdoAdoAdo (dismissed):

AdoAdoAdo (dismissed):

andreibancioiu (approved):

AdoAdoAdo (approved):

PR #9, opened 2021-06-23
En 10293/implement initial APIs

mariusmihaic:
- Account class now has additional constructor whith address, balance, nonce
- Added transaction status class
- Added wrapper for ERD API generic responses: `ErdGenericApiResponse`
- HTTP Client can post message with different content types
- Added missing file `pem_input_wrapper.h `file, which I forgot to add it to repository in past PRs

Implemented ProxyProvider with the following functionalities:
- `getAccount`
- `send` transaction
- `getTransactionStatus`


cli/inputhandler/wrappers/pem_input_wrapper.h
```diff
@@ -0,0 +1,29 @@
...
+
+    const std::string &getPemFilePath() const
+    {
+        return getInputData().at(ARGS_TX_IDX_PEM_INPUT_FILE);
+    }
+};
+}
+}
+
+#endif
```
miiu96: add empty line

include/account/account.h
```diff
@@ -4,16 +4,27 @@
...
+#endif
+
+#define DEFAULT_BALANCE "0"
+
 class Account
 {
 public:
     explicit Account(Address address);
 
+    explicit Account(Address address, std::string balance, uint64_t const &nonce);
```
andreibancioiu, line 18: 👍

andreibancioiu: Perhaps in the future we can implement a specialized class for handling the amount of tokens and use it instead of `std::string balance`.

include/account/account.h
```diff
@@ -4,16 +4,27 @@
...
+
     const Address& getAddress() const;
 
     const std::string& getBalance() const;
 
+    const uint64_t& getNonce() const;
```
andreibancioiu, line 24: `incrementNonce` will also be quite useful (e.g. see erdpy and erdjs).

include/provider/data/data_transaction.h
```diff
@@ -0,0 +1,33 @@
+#ifndef ERD_PROXY_DATA_TRANSACTION_H
+#define ERD_PROXY_DATA_TRANSACTION_H
+
+#include <string>
+
+struct TransactionHash
```
andreibancioiu, line 6: Perhaps add a comment which explains that the `hash` member is in hex?

include/provider/data/data_transaction.h
```diff
@@ -0,0 +1,33 @@
...
+{
+    std::string hash;
+};
+
+class TransactionStatus
```
andreibancioiu, line 14: Perhaps also add a comment which references the `erdjs` implementation (e.g. "Also see ...")?

include/provider/proxyprovider.h
```diff
@@ -0,0 +1,26 @@
+#ifndef ERD_PROXY_PROVIDER_H
+#define ERD_PROXY_PROVIDER_H
+
+#include "data/ext.h"
+#include "account/account.h"
+#include "account/address.h"
+#include "transaction/transaction.h"
+
+class ProxyProvider
```
andreibancioiu, line 9: 👍

src/provider/apiresponse.h
```diff
@@ -0,0 +1,54 @@
+#ifndef ERD_PROXY_API_RESPONSE_H
+#define ERD_PROXY_API_RESPONSE_H
+
+#include "json/json.hpp"
+#include "../utils/errors.h"
+
+class ErdGenericApiResponse
+{
+public:
+    explicit ErdGenericApiResponse(std::string rawData)
```
andreibancioiu: Usually, for this kind of object construction, we define a static function called `parse...()` that returns the object. Just a suggestion - fine as it is, though.

src/provider/apiresponse.h
```diff
@@ -0,0 +1,54 @@
...
+    {
+        try
+        {
+            m_response = nlohmann::json::parse(rawData);
+        }
+        catch (...)
+        {
+            throw std::invalid_argument(ERROR_MSG_JSON_SERIALIZED + rawData);
```
andreibancioiu, line 31: 👍

src/provider/apiresponse.h
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    std::string getError() const
+    {
+        std::string ret;
+
+        if (m_response.contains("error"))
+            if(m_response["error"] != "")
+                ret = m_response["error"];
```
andreibancioiu, line 56: Just return, don't define & assign? Fine as it is, though.

About code formatting - aren't the curly braces mandatory in our set of rules (here, and below, in other functions)?

src/provider/apiresponse.h
```diff
@@ -0,0 +1,54 @@
...
+
+       return ret;
+    }
+
+    std::string getCode() const
+    {
+        if (!m_response.contains("code"))
```
andreibancioiu: Here, and below - we can define an extra function called `requireAttribute()` and remove bits of duplication. Not mandatory.

src/provider/data/data_transaction.cpp
```diff
@@ -0,0 +1,34 @@
+#include "provider/data/data_transaction.h"
+
+TransactionStatus::TransactionStatus(std::string status) :
+    m_status(std::move(status)){}
+
+bool TransactionStatus::operator ==(TransactionStatus const& txStatus)
```
andreibancioiu, line 6: 👍

src/provider/data/data_transaction.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+    return this->m_status == txStatus.m_status;
+}
+
+bool TransactionStatus::isPending() const
+{
+    return (m_status == "received") || (m_status == "pending") || (m_status == "partially-executed");
```
andreibancioiu: I think `partially-executed` is now obsolete - @miiu96 can confirm. But it's fine either way (I think erdjs handles it as well).

src/provider/proxyprovider.cpp
```diff
@@ -0,0 +1,86 @@
+#include "provider/proxyprovider.h"
+#include "apiresponse.h"
+#include "httpwrapper.h"
+
+namespace internal
+{
+    bool isRequestSuccessful(ErdGenericApiResponse const &response)
```
andreibancioiu: Not sure, why isn't this a member function (improve cohesion)?

src/provider/apiresponse.h
```diff
@@ -0,0 +1,54 @@
...
+        return m_response["code"];
+    }
+
+    template <class T>
+    T getData() const
+    {
+        if (!m_response.contains("data"))
+            throw std::invalid_argument(ERROR_MSG_JSON_KEY_NOT_FOUND + "data");
+
+        return m_response["data"];
```
andreibancioiu, line 75: So here, the indexing operator `[...]` is generic as well, right? 

Not sure, is it possible to rewrite `getData()` to include the logic in `getDataIfValid()` as well? And perhaps also rename it to `extractData()`? `getPayload()`?

Or perhaps move the logic from `getDataIfValid` to the constructor (`parseResponse() / parseJson()`)?

How about this piece of code (just an idea)?

```
auto response = ErdGenericApiResponse.parseJson("...");
response.raiseOnError();
auto foo = response.getPayload();
```

src/provider/proxyprovider.cpp
```diff
@@ -0,0 +1,86 @@
...
+    m_url(std::move(url)) {}
+
+Account ProxyProvider::getAccount(Address const &address)
+{
+    wrapper::http::Client client(m_url);
+    wrapper::http::Result const result = client.get("/address/" + address.getBech32Address());
+
+    auto data = internal::getDataIfValid(result);
+
+    if (!data.contains("account"))
```
andreibancioiu: `requireAttribute()`, utility function? However, I think these assertions are a bit redundant. If the API responded with success, then it didn't lie :)

src/provider/proxyprovider.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+TransactionHash ProxyProvider::send(Transaction const &transaction)
+{
+    wrapper::http::Client client(m_url);
+    wrapper::http::Result const result = client.post("/transaction/send", transaction.serialize(), wrapper::http::applicationJson);
+
+    auto data = internal::getDataIfValid(result);
+
+    if (!data.contains("txHash"))
```
andreibancioiu: As above. Doesn't `data["txHash"];` throw a good-enough exception if the key is missing, anyway?

src/wrappers/httpwrapper.h
```diff
@@ -32,17 +40,37 @@ class Client
...
     }
 
 private:
+    std::string getContentType(ContentType const &contentType) const
+    {
+        switch (contentType)
+        {
+            case textPlain:
+            {
+                return CONTENT_TYPE_PLAIN_TEXT;
```
andreibancioiu: Since this CPP SDK will be mostly used to interact with the Elrond APIs, which are 100% JSON-based, perhaps adjust the default values? (here and above)

src/wrappers/httpwrapper.h
```diff
@@ -32,17 +40,37 @@ class Client
...
+                break;
+            }
+            case applicationJson:
+            {
+                return CONTENT_TYPE_JSON;
```
andreibancioiu, line 57: Just return, and no break?

tests/test_src/test_provider/test_apiresponse.cpp
```diff
@@ -0,0 +1,100 @@
...
+TEST(ErdGenericApiResponse, getters_allExist)
+{
+    ErdGenericApiResponse apiResponse(R"({"data":{"account":{"address":"erd1qq"}},"error":"some error","code":"some code"})");
+
+    nlohmann::json expectedData;
+    expectedData["account"]["address"] = "erd1qq";
+
+    EXPECT_EQ(apiResponse.getCode(), "some code");
+    EXPECT_EQ(apiResponse.getError(), "some error");
+    EXPECT_EQ(apiResponse.getData<nlohmann::json>(), expectedData);
```
andreibancioiu, line 34: For a more beautiful test, can you define a struct instead of using `nlohmann::json`,  and then call `getData<MyStruct>()`?

Or just rewrite the assertion to not use `nlohmann::json` at all - here and in the tests below? (optionally).

tests/test_src/test_provider/test_proxyprovider.cpp
```diff
@@ -0,0 +1,111 @@
+#include "gtest/gtest.h"
+
+#include "provider/proxyprovider.h"
+#include "filehandler/pemreader.h"
+#include "utils/errors.h"
+
+
+TEST(ProxyProvider, getAccount)
```
andreibancioiu, line 10: 👍

include/account/account.h
mariusmihaic: Yes, maybe something like here (a Balance class) in erdjs: https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/5c10d7d69dbc424464909398d0874684b59cb2c4/src/balance.ts#L27
For now, this is not a priority yet, but should be added in ERDCPP as well

include/account/account.h
mariusmihaic: Good point. Implement it + added gtests as well

include/provider/data/data_transaction.h
mariusmihaic: Added a comment for it

include/provider/data/data_transaction.h
mariusmihaic: Added comment

include/provider/proxyprovider.h
mariusmihaic: Thanks !

src/provider/apiresponse.h
mariusmihaic: > Just return, don't define & assign?

I've noticed that for different API urls, we get different responses.
E.g.: 
- When sending a valid transaction with URL = https://testnet-api.elrond.com , we **don't** receive **error** attribute,
- When sending a valid transaction with URL = https://testnet-gateway.elrond.com , we receive _**empty**_ **error** attribute.

This is the reason why the code looks like that.

> About code formatting - aren't the curly braces mandatory in our set of rules (here, and below, in other functions)?

I guess you are right, I've added them !

src/provider/apiresponse.h
mariusmihaic: Added it !

src/provider/data/data_transaction.cpp
mariusmihaic: Thanks ! :)

src/provider/data/data_transaction.cpp
AdoAdoAdo: also "received" is obsolete
so only "pending" status should still be checked

include/provider/proxyprovider.h
```diff
@@ -0,0 +1,26 @@
...
+
+    TransactionHash send(Transaction const &transaction);
+
+    TransactionStatus getTransactionStatus(std::string const &txHash);
+
+private:
+    std::string m_url;
+};
+
+#endif //ERD_API_H
```
AdoAdoAdo: wrong comment?
ERD_PROXY_PROVIDER_H

src/provider/apiresponse.h
```diff
@@ -0,0 +1,54 @@
...
+#include "../utils/errors.h"
+
+class ErdGenericApiResponse
+{
+public:
+    explicit ErdGenericApiResponse(std::string rawData)
+    {
+        try
+        {
+            m_response = nlohmann::json::parse(rawData);
```
AdoAdoAdo, line 27: can we use our json wrapper instead in this class?

src/provider/apiresponse.h
mariusmihaic: I've applied your suggestion and created a function that parses `wrapper::http::Result` and returns an `ErdGenericApiResponse`:
`ErdGenericApiResponse parse(wrapper::http::Result const &res)`

src/provider/apiresponse.h
mariusmihaic: Thanks ! :)

src/provider/data/data_transaction.cpp
miiu96: You can remove (m_status == "received") and (m_status == "partially-executed") because proxy will respond only with the status `pending`.

src/provider/proxyprovider.cpp
mariusmihaic: I've removed that function from internal namespace anyway.
However, there were two reasons for this:

1. improve cohesion -> YES. That function did not use any internal member function, was rather a helper
2. **The most important reason:** I did not want to include`apiresponse.h` in `proxyprovider.h`. This way, it is only used in `proxyprovider.cpp`, so CPP SDK users don't have to include internal files for their APIs (`apiresponse.h`)

src/wrappers/httpwrapper.h
```diff
@@ -32,17 +40,37 @@ class Client
...
 
 private:
+    std::string getContentType(ContentType const &contentType) const
+    {
+        switch (contentType)
+        {
+            case textPlain:
+            {
+                return CONTENT_TYPE_PLAIN_TEXT;
+                break;
```
AdoAdoAdo: the break after return is unreachable code.

either return directly without break, or have a single exit point in the function at the end where you return the content type, and on the case set the  content type to be returned and break.

src/provider/data/data_transaction.cpp
mariusmihaic: Removed "received" and "partially-executed".
Thanks for heads up !

src/provider/apiresponse.h
mariusmihaic: > So here, the indexing operator [...] is generic as well, right?

In `ErdGenericApiResponse` -> YES, it is generic and returns a template class.
However, in `ProxyProvider` we always expect data (if valid) to be a json.

> How about this piece of code (just an idea)?

I've applied your suggestion.
However, I've decided to include those 3 lines of code in a function to avoid duplicate code.
So now, we have this internal function, with your proposed functionality:
```
nlohmann::json getPayLoad(wrapper::http::Result const &res)
    {
        ErdGenericApiResponse response = internal::parse(res);
        response.checkSuccessfulOperation();

        return response.getData<nlohmann::json>();
    }
```

src/wrappers/httpwrapper.h
mariusmihaic: Removed unreachable break !

src/wrappers/httpwrapper.h
mariusmihaic: Good point. Changed default content type to JSON-based here and above !

tests/test_src/test_provider/test_proxyprovider.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    EXPECT_FALSE(account.getAddress().getBech32Address().empty());
+    EXPECT_FALSE(account.getBalance().empty());
+    EXPECT_FALSE(account.getBalance() == DEFAULT_BALANCE);
+    EXPECT_FALSE(account.getNonce() == 0);
+}
+
+TEST(ProxyProvider, send_validTx)
+{
+    ProxyProvider proxy("https://testnet-gateway.elrond.com");
```
AdoAdoAdo, line 25: can we have this kind of tests under some some flag or skipped by default?
for this test to pass there are also some external preconditions:
1. testnet is up and running
2. used account still has enough funds
3. there are no concurrent transactions using the same account on testnet
etc

we can have the tests disabled by default and add some comment on the preconditions that need to be satisfied if the tests are enabled.

src/provider/proxyprovider.cpp
mariusmihaic: After yesterday's discussion, we decided to check anyway for these fields. I've added `requireAttribute` utility function, however.

src/provider/proxyprovider.cpp
mariusmihaic: As discussed yesterday, we will keep it as it is

tests/test_src/test_provider/test_apiresponse.cpp
mariusmihaic: As discussed yesterday, we will keep it as it is

include/provider/proxyprovider.h
mariusmihaic: Yes, forgot it there, fixed it

src/provider/apiresponse.h
mariusmihaic: As discussed yesterday, we'll keep that one for now. I'll try to create another wrapper in the future.

src/wrappers/httpwrapper.h
mariusmihaic: Yes, I've missed it :(
I've fixed it now, I will simply return content without break.

tests/test_src/test_provider/test_proxyprovider.cpp
mariusmihaic: I've added a flag for this.
Disabling tests with prefix : `DISABLED_` would not have been the best solution, since it requires manually setting this flag for each test.
Instead I've added a `#define FLAG OFF` and used `#if FLAG` for all tests

src/provider/apiresponse.h
andreibancioiu: They are different APIs. For example, in erdjs, there are two separate "providers" (API client wrappers): one for the General API, one for the Gateway (Proxy API). Perhaps we should have two separate providers in erdcpp as well, at some point?

andreibancioiu (approved):

src/provider/proxyprovider.cpp
```diff
@@ -40,14 +31,11 @@ Account ProxyProvider::getAccount(Address const &address)
     wrapper::http::Client client(m_url);
     wrapper::http::Result const result = client.get("/address/" + address.getBech32Address());
 
-    auto data = internal::getDataIfValid(result);
+    auto data = internal::getPayLoad(result);
```
AdoAdoAdo, line 34: typo getPayload instead of getPayLoad

AdoAdoAdo (approved):

PR #10, opened 2021-06-25
EN-10079/Message verify

mariusmihaic:
Added crypto library function wrappers for:
- getPublicKey (from secret key)
- verify (signature, message, public key)

In `Signer`, added verify signed message by:

- signer
- public key
- public address

### DIFF
--- a/include/transaction/signer.h
+++ b/include/transaction/signer.h
@@ -10,6 +10,10 @@ public:
 
     std::string getSignature(std::string const &message) const;
 
+    bool verify(std::string const &signature, std::string const &message) const;
+
+    static bool verify(std::string const &signature, std::string const &message, bytes const &publicKey);
+
 private:
     bytes m_sk;
 };
--- a/include/transaction/signer.h
+++ b/include/transaction/signer.h
@@ -2,6 +2,7 @@
 #define ERD_SIGNER_H
 
 #include "internal/internal.h"
+#include "account/address.h"
 
 class Signer
 {
@@ -13,6 +14,8 @@ public:
     bool verify(std::string const &signature, std::string const &message) const;
 
     static bool verify(std::string const &signature, std::string const &message, bytes const &publicKey);
+
+    static bool verify(std::string const &signature, std::string const &message, Address const &address);
 
 private:
     bytes m_sk;
--- a/tests/test_src/test_transaction/test_transaction.cpp
+++ b/tests/test_src/test_transaction/test_transaction.cpp
@@ -4,6 +4,7 @@
 #include "utils/errors.h"
 #include "transaction/signer.h"
 #include "transaction/transaction.h"
+#include "wrappers/cryptosignwrapper.h"
 
 
 class SignerConstructorFixture : public ::testing::Test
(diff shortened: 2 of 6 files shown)


tests/test_src/test_wrappers/test_wrappers.cpp
```diff
@@ -151,6 +152,25 @@ TEST(CryptoWrapper, getSignature)
...
+TEST(CryptoWrapper, getPublicKey)
+{
+    bytes const seed = util::hexToBytes("413f42575f7f26fad3317a778771212fdb80245850981e48b58a4f25e344e8f9");
+    bytes const secretKey =  wrapper::crypto::getSecretKey(seed);
+    bytes const expectedPublicKey = util::hexToBytes("0139472eff6886771a982f3083da5d421f24c29181e63888228dc81ca60d69e1");
+
+    EXPECT_EQ(wrapper::crypto::getPublicKey(secretKey), expectedPublicKey);
+}
+
+TEST(CryptoWrapper, verify)
```
andreibancioiu, line 169: Add a comment that contains a link to the reference tests (e.g. erdpy, erdjs, elrond-go)?

tests/test_src/test_transaction/test_transaction.cpp
```diff
@@ -49,6 +50,39 @@ TEST(Signer, getSignature)
     EXPECT_EQ(util::stringToHex(signature), expectedSignature);
 }
 
+TEST(Signer, verify_signature_by_signer)
```
andreibancioiu, line 58: Please add a link to the source code of the reference tests (erdjs, elrond-go).

include/transaction/signer.h
```diff
@@ -10,6 +11,12 @@ class Signer
 
     std::string getSignature(std::string const &message) const;
 
+    bool verify(std::string const &signature, std::string const &message) const;
```
andreibancioiu, line 14: Nice! 

Is this in preparation for [this feature](https://github.com/ElrondNetwork/elrond-sdk/pull/215)?

Now I realise that our `signer.h` shouldn't necessarily be under `include/transaction/`. Perhaps better under `include/wallet/`. However, OK for the moment.

tests/test_src/test_wrappers/test_wrappers.cpp
mariusmihaic: Added used reference links here and for transaction tests are well.

tests/test_src/test_transaction/test_transaction.cpp
mariusmihaic: Added reference links here and for transaction tests are well.

include/transaction/signer.h
mariusmihaic: > Is this in preparation for this feature?

I took a peek at that PR. However, I think that one is for signing/verifying messages with keccak ? (prefix + msgLength + msg). I think we will also need to implement that functionality in ERDCPP as well.
The reason why we implemented this is because it was required for a first version of ERDCPP to be able to:
> verify message (offchain)


> Now I realise that our signer.h shouldn't necessarily be under include/transaction/. Perhaps better under include/wallet/. However, OK for the moment.

Very good point. However, we don't have any **wallet** yet. For current version/structure, it was meant to be used only to sign transactions and messages, but as this SDK grows, we will change it

AdoAdoAdo: yes for the message signing we will need the keccak hashing, so looking for a C++ library.
this PR can go as it is, the prefix concatenation + hashing can be added in another PR once we decide on the library, which then makes use of the existing signing & verification methods.

andreibancioiu (approved):

AdoAdoAdo (approved):

PR #11, opened 2021-06-28
FEAT: Add APIs for getESDT tokens

mariusmihaic:
Added 2 new APIs in ProxyProvider:

- `getESDTTokenBalance`: returns token balance for given address
- `getAllESDTTokenBalances`: returns all token balances for given address


src/provider/proxyprovider.cpp
```diff
@@ -70,3 +70,45 @@ TransactionStatus ProxyProvider::getTransactionStatus(std::string const &txHash)
 
     return TransactionStatus(txStatus);
 }
+
+std::string ProxyProvider::getESDTTokenBalance(Address const &address, std::string const &token) const
+{
+    wrapper::http::Client client(m_url);
+    wrapper::http::Result const result = client.get("/address/" + address.getBech32Address() + "/esdt/" + token);
```
andreibancioiu, line 77: Is string interpolation possible (here, below etc.)?

https://stackoverflow.com/a/57128991/1475331

src/provider/proxyprovider.cpp
```diff
@@ -70,3 +70,45 @@ TransactionStatus ProxyProvider::getTransactionStatus(std::string const &txHash)
...
+
+    auto esdts = data["esdts"];
+
+    std::map<std::string, std::string> ret;
+    std::string esdt;
+    std::string balance;
+
+    for (const auto& it : esdts.items())
+    {
+        utility::requireAttribute(it.value(), "balance");
```
andreibancioiu, line 106: Not sure, can we have some functionality such as:

```
utility::getAttribute(someData, attributeName, fallbackValueIfAttributeIsMissingOrNull)
```

Not in this PR, but in a future one, if it proves useful (e.g. in this case?)

tests/test_src/test_provider/test_proxyprovider.cpp
```diff
@@ -112,4 +112,36 @@ TEST(ProxyProvider, getTransactionStatus_invalidHash)
...
+    EXPECT_TRUE(esdts.empty());
+}
+
+TEST(ProxyProvider, getAllESDTokenBalances_multipleTokens)
+{
+    ProxyProvider proxy("https://testnet-gateway.elrond.com");
+    Address const address("erd1nqtv8gdsf55xj9eg0wc8ar6ml46kpld2c86aq670mxgcf49sduzqaeyngx");
+    auto const esdts = proxy.getAllESDTTokenBalances(address);
+
+    EXPECT_TRUE(esdts.find("00040-4c4d18") != esdts.end());
```
andreibancioiu, line 140: Tricky test. I don't think it will pass after a testnet reset or when `erd1nqtv...` transfers its tokens.

Perhaps (in a future PR), an option would be to make the integration tests depend on a local testnet (erdpy), and also add some setup / initialization steps before running the actual tests (e.g. steps such as creating some ESDT tokens etc.).

andreibancioiu (dismissed):

AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

src/provider/proxyprovider.cpp
mariusmihaic: Unfortunately, it is only possible from C++20 (I've also tried it myself) :(
We are using C++14...

src/provider/proxyprovider.cpp
mariusmihaic: It would be useful as a fallback, but in this case I don't think we want that.

Suppose that somehow we get a ProxyResponse that doesn't contain a field, but we fill it with default value (e.g. default balance = "0"). For those who use the response, since we don't throw any exception, they will use that balance thinking it is correct.

tests/test_src/test_provider/test_proxyprovider.cpp
mariusmihaic: Totally agree, these tests are not safe. That's also the reason why these tests are deactivated by `HTTP_PRECONDITIONS` flag. Your suggestion is very good. Also had a discussion with @AdoAdoAdo and we should use a local testnet in the future.

andreibancioiu (approved):

PR #13, opened 2021-07-01
En 10239/transfer esdt tokens

mariusmihaic:
- Added dependency to GNU GMP library to handle big int 
- Added `BigUInt` class
- Added `SCArguments` class to handle smart contract arguments call for transactions
- Added function to prepare ESDT token transfer with or without SC function & parameters
- Added gtests


AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

include/internal/biguint.h
```diff
@@ -0,0 +1,19 @@
+#ifndef ERD_BIG_UINT_H
+#define ERD_BIG_UINT_H
+
+#include <string>
+
+class BigUInt
+{
+public:
+    explicit BigUInt(std::string value);
```
andreibancioiu, line 9: 👍

include/transaction/scarguments.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "internal/biguint.h"
+#include "account/address.h"
+
+class SCArguments
+{
+public:
+    explicit SCArguments();
+
+    void add(std::string const &arg);
```
andreibancioiu, line 14: 👍

src/transaction/scarguments.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    m_args.emplace_back(val);
+}
+
+void SCArguments::add(Address const &arg)
+{
+    bytes const pk = arg.getPublicKey();
+
+    std::string pkHex(pk.begin(), pk.end());
+    pkHex = util::stringToHex(pkHex);
```
andreibancioiu, line 27: Perhaps in the future we can add `toHex` (`toPubKey`) to `Address` class as well.

src/transaction/scarguments.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return m_args.empty();
+}
+
+std::string SCArguments::asOnData() const
+{
+    std::string ret;
+
+    for (auto const &arg : m_args)
+    {
+        ret += "@" + arg;
```
andreibancioiu, line 43: So, `asOnData()` always returns having a leading `@`, right?

tests/test_src/test_transaction/test_transaction.cpp
```diff
@@ -495,3 +496,210 @@ TEST_F(TransactionSerializeFixture, serialize_missingFields)
     tx.m_receiver = nullptr;
     expectSerializeException<std::invalid_argument>(tx, ERROR_MSG_RECEIVER);
 }
+
+TEST(SCArguments, add_emtpy_asOnData)
```
andreibancioiu, line 500: 👍

andreibancioiu (approved):

include/internal/biguint.h
mariusmihaic: Thanks ! :+1:

include/transaction/scarguments.h
mariusmihaic: Thanks ! :+1:

tests/test_src/test_transaction/test_transaction.cpp
mariusmihaic: Thanks ! :+1:

src/transaction/scarguments.cpp
mariusmihaic: Yes, it will always return the params preceded with `@`

src/transaction/scarguments.cpp
mariusmihaic: Maybe we could even:

- Change bytes typedef to be a class
- Add `toHex` method to it ?

``` c++
class Bytes(std::string input)
public: 
std::string toHex();
```

PR #14, opened 2021-07-01
EN-10322/ESDT issuance

mariusmihaic:
- Renamed esdt transfer related defines form `esdt.h`
- Added `ESDTProperties` struct. In the future, we could have a `ESDT` class/struct, which will inherit this struct.
- Added `!=` and `==` comparison operators for `ESDTProperties`
- Added functionality to issue esdt token with or without default esdt properties
- Added `GenericProxyProviderTxFixture` gtest fixture to send transactions using proxy and a pem file
- Added `TestnetProxyProviderTxFixture` gtest fixture which instantiates `GenericProxyProviderTxFixture` with https://testnet-gateway.elrond.com
-  Added `DevnetProxyProviderTxFixture` gtest fixture which instantiates `GenericProxyProviderTxFixture` with https://devnet-gateway.elrond.com


tests/test_src/test_transaction/test_transaction.cpp
```diff
@@ -703,3 +703,64 @@ TEST_P(PrepareEsdtTransferData, noFunction)
     EXPECT_EQ(tx.m_gasLimit, currParam.gasLimitAfterPrep);
     EXPECT_EQ(txDataAfterPrep, currParam.dataAfterPrep);
 }
+
+TEST(ESDTProperties, comparisonOperators)
```
andreibancioiu, line 707: Perhaps in the future we might want to extract these ESDT-specific tests to a separate file.

tests/test_src/test_provider/test_proxyprovider.cpp
```diff
@@ -144,4 +81,181 @@ TEST(ProxyProvider, getAllESDTokenBalances_multipleTokens)
     EXPECT_FALSE(esdts.at("0009O-8742a4").empty());
 }
 
+class GenericProxyProviderTxFixture : public ::testing::Test
```
andreibancioiu, line 84: 👍 Nice refactoring for integration tests.

include/transaction/esdt.h
```diff
@@ -4,15 +4,43 @@
...
-#define NO_PARAMETERS SCArguments()
-
-#define ESDT_GAS_LIMIT_NO_FUNCTION 500000
 #define ESDT_TRANSFER_PREFIX std::string("ESDTTransfer")
+#define ESDT_TRANSFER_GAS_LIMIT_NO_FUNCTION 500000
+#define ESDT_TRANSFER_NO_FUNCTION std::string()
+#define ESDT_TRANSFER_NO_ARGUMENTS SCArguments()
+
+#define ESDT_ISSUANCE_PREFIX std::string("issue")
+#define ESDT_ISSUANCE_VALUE "50000000000000000" //(0.05 EGLD)
```
andreibancioiu, line 13: Shouldn't this be 5?

include/transaction/esdt.h
```diff
@@ -4,15 +4,43 @@
...
+#define ESDT_ISSUANCE_GAS_LIMIT 60000000
```
andreibancioiu, line 14: https://github.com/ElrondNetwork/elrond-config-testnet/blob/master/gasSchedules/gasScheduleV3.toml#L28

Perhaps rewrite the constant as 50000000 + 10000000, or perhaps compute the exact gas limit after building the transaction data?

src/transaction/esdt.cpp
```diff
@@ -1,6 +1,35 @@
 #include "hex.h"
 #include "transaction/esdt.h"
 
+namespace internal
+{
+
+std::string ESDTPropertyField(std::string const &property, bool const &val)
```
andreibancioiu, line 7: `Flag` instead of `PropertyField`? (a less generic name) - just a suggestion.

Later edit: Oh, now I see the correct name is `property` (as in docs). Can stay as it is then.

src/transaction/esdt.cpp
```diff
@@ -1,6 +1,35 @@
...
+{
+    std::string const value = val ? "true" : "false";
+
+    return "@" + util::stringToHex(property) + "@" + util::stringToHex(value);
+}
+
+}
+
+bool operator==(ESDTProperties const &lhs, ESDTProperties const &rhs)
```
andreibancioiu, line 16: 👍

src/transaction/esdt.cpp
```diff
@@ -28,4 +57,34 @@ void prepareTransactionForESDTTransfer(Transaction &transaction,
...
 
+void prepareTransactionForESDTIssuance(Transaction &transaction,
+                                      std::string const &token,
+                                      std::string const &ticker,
+                                      std::string const &initialSupply,
+                                      std::string const &noOfDecimals,
+                                      ESDTProperties const &esdtProperties)
+{
+    transaction.m_value = ESDT_ISSUANCE_VALUE;
+    transaction.m_gasLimit = ESDT_ISSUANCE_GAS_LIMIT;
```
andreibancioiu, line 68: Perhaps actually compute the gas limit?

andreibancioiu (dismissed):

tests/test_src/test_transaction/test_transaction.cpp
mariusmihaic: Yes. It would be worth splitting them in the future as more tests are going to be added.

tests/test_src/test_provider/test_proxyprovider.cpp
mariusmihaic: Thank you ! :smile:

include/transaction/esdt.h
AdoAdoAdo: it was changed to 0.05 egld

mariusmihaic: Info is from here: https://docs.elrond.com/developers/esdt-tokens/. It seemns to be 0.05 EGLD

src/transaction/esdt.cpp
mariusmihaic: Thanks ! :)

include/transaction/esdt.h
mariusmihaic: Based on docs :https://docs.elrond.com/developers/esdt-tokens/#issuance-of-fungible-esdt-tokens it just states that it should be 60000000, but in the link you attached it seems to be just 50000000. That seems odd.

> perhaps compute the exact gas limit after building the transaction data?

I think this is the right way to go. We should add function to compute transaction gas limit in the future.

src/transaction/esdt.cpp
mariusmihaic: Yes, I've named it following docs naming

src/transaction/esdt.cpp
mariusmihaic: Yes, that would be way better.
We shall add a functionality for it in the future.

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #15, opened 2021-07-08
En 10081/accept cli arguments any order

mariusmihaic:
- Changed current CLI implementation, using `cxxoptions.h` library (header-only file, **should NOT be reviewed** )
- Deleted old code which handled CLI, replaced logic with classes from library
- Binary can accept CLI arguments in any order
- Binary CLI expects arguments organized on 3 levels:
1. command (e.g.: **pem** ; **transaction**)
2. sub-command (e.g.: pem **load** ; transaction **new**)
3. arguments (e.g.: pem load **--file="someFile"** ; transaction new **--nonce=4 --value=100000...** )

- Added `options` class which holds all valid options that CLI can parse
- Changed `ArgHandler` class to holds all `options` (as internal class member) + added public function to parse arguments
- Refactored CLI input wrappers
- Refactored gtests


cli/inputhandler/arghandler.cpp
```diff
@@ -1,158 +1,101 @@
...
-    // TODO: No magic numbers/strings
-    if ((argCount() == 1) && isCmdGroup("help"))
+    if (isCmd("help") && argc == 2)
     {
+        helpMsg = m_options.help();
         reqType = help;
     }
-    else if ((argCount() == 3) && isCmdGroup("pem") && isSubCmd(1U, "load") &&
-             (checkAndSetUserInput<std::string>(2U, "--file=", userInputs, ARGS_TX_IDX_PEM_INPUT_FILE, ERROR_PEM_INPUT_FILE)))
+    else if (isCmd("pem") && isSubCmd("help") && argc == 3)
```
AdoAdoAdo: please have isSubCmd("--help") || isSubCmd("-h") instead of isSubCmd("help"). Same for isCmd("help")

cli/inputhandler/options.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+cxxopts::Options CLIOptions::pem() const
+{
+    return m_optionsPem;
+}
+
+std::string CLIOptions::help() const
+{
+    return helpTx() + "\n" + helpPem();
```
AdoAdoAdo: can you adapt a bit the help and add some more details before jumping to subcommand options?
binary command subcommand [OPTIONS...]

cli/commandhandler.h
```diff
@@ -32,103 +32,56 @@ void signTransaction(Transaction &transaction, bytes const& seed)
...
+    PemFileReader pemReader(transactionInputWrapper.getInputFile());
 
     Transaction transaction = internal::createTransaction(transactionInputWrapper, pemReader.getAddress());
     internal::signTransaction(transaction,pemReader.getSeed());
 
     jsonFile.writeDataToFile(transaction.serialize());
 }
 
-void handleRequest(ih::RequestedCmd const &requestedCmd)
+void handleRequest(ih::ArgParsedResult const &parsedRes)
```
andreibancioiu: Why `parsedRes` and not `parsedResult`?

cli/inputhandler/options.cpp
```diff
@@ -0,0 +1,64 @@
...
+            ("chainID", "Chain identifier", cxxopts::value<std::string>()->default_value("T"))
+            ("version", "Transaction version", cxxopts::value<uint64_t>()->default_value("1"))
+            ("options", "Transaction options", cxxopts::value<uint32_t>()->default_value("0"))
+            ("pem", "Pem keyfile", cxxopts::value<std::string>())
+            ("outfile", "Json file where the output will be stored", cxxopts::value<std::string>());
+}
+
+void CLIOptions::initOptionsPem()
+{
+    m_optionsPem.add_options("load") // pem load group
```
andreibancioiu: Not sure what is the purpose of this command.

cli/inputhandler/options.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+void CLIOptions::initOptions()
+{
+    initOptionsTx();
+    initOptionsPem();
+}
+
+void CLIOptions::initOptionsTx()
+{
+    m_optionsTx.add_options("new") // transaction new group
```
andreibancioiu, line 32: Not sure, are there use cases for the CLI, in general, or is it for demonstrating the SDK usage?

cli/inputhandler/wrappers/transaction_input_wrapper.h
```diff
@@ -46,54 +43,49 @@ namespace wrapper
...
     {
-        return Address(getInputData().at(ARGS_TX_IDX_RECEIVER));
+        std::string const address = getInputData()["receiver"].as<std::string>();
+        return Address(address);
     }
 
     std::shared_ptr<bytes> getReceiverName() const
     {
-        return internal::getUserInputIfExists<bytes>(getInputData(), ARGS_TX_IDX_RECEIVER_NAME, m_containsReceiverName);
+        return internal::getUserInputIfExists<bytes>(getInputData(), "receiver-name");
```
andreibancioiu, line 68: Not sure, is this correct? Above you had `sender-username` & `receiver-username`. I think we can define the keys as constants, but it might be an overkill, so it can stay as it is :)

andreibancioiu (commented):
Very nice & clean - not sure though if a new CLI is needed / required / desired. But it's perfect for demonstrating the power of the CPP SDK, though.

cli/inputhandler/options.cpp
AdoAdoAdo: right now I think it is just for demonstrating the SDK.

cli/commandhandler.h
mariusmihaic: Probably rushed to name the variable, will rename it to make it more verbose

cli/inputhandler/options.cpp
mariusmihaic: Actually, you are right.
One of the first requirements of this SDK was to load a pem file to check if it was valid. I kept carrying this implementation, but right now it just does **not** make sense to keep it anymore.
When you try to create a new transaction or issue/transfer ESDTs, we check the validity of the pem file anyway.

I will **DELETE** it

cli/inputhandler/options.cpp
mariusmihaic: Yes, just for some functionality demonstration

cli/inputhandler/wrappers/transaction_input_wrapper.h
mariusmihaic: Good finding. It was wrong, I had it mixed it up, so I fixed it.

> I think we can define the keys as constants, but it might be an overkill, so it can stay as it is :)

This was also my first though: make them constants. But then I realized that we will only have few CLI functionalities, so might be indeed an overkill.

cli/inputhandler/arghandler.cpp
mariusmihaic: Yes. Added `-h` and `--help` as well

cli/inputhandler/options.cpp
mariusmihaic: I've adapted it to look something like this: 
image

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #16, opened 2021-07-09
En 10335/add cli esdt options

mariusmihaic:
- Added function to handle ESDT Issue
- Added function to handle ESDT Transfer
- Added external library `external/toml/cpptoml.h` - should **NOT** be reviewed !
- Added toml network configuration
- Added `NetworkConfig` class
- Added options for esdt : issue + transfer
- Rename cli binary from main -> erdcpp

### DIFF
--- a/cli/commandhandler.h
+++ b/cli/commandhandler.h
@@ -122,6 +122,7 @@ void handleTransferESDT(cxxopts::ParseResult const &result)
 
     auto const nonce = result["nonce"].as<uint64_t>();
     auto const gasPrice = result["gas-price"].as<uint64_t>();
+    auto const gasLimit = result["gas-limit"].as<uint64_t>();
     auto const receiverAdr = result["receiver"].as<std::string>();
     auto const token = result["token"].as<std::string>();
     auto const value = result["value"].as<std::string>();
@@ -140,6 +141,7 @@ void handleTransferESDT(cxxopts::ParseResult const &result)
     tx.m_sender = std::make_shared<Address>(sender);
     tx.m_receiver = std::make_shared<Address>(receiver);
     tx.m_gasPrice = gasPrice;
+    tx.m_gasLimit = gasLimit;
     tx.m_chainID = config.chainID;
 
     prepareTransactionForESDTTransfer(tx, token, function);
--- a/cli/commandhandler.h
+++ b/cli/commandhandler.h
@@ -24,9 +24,11 @@ Transaction createTransaction(ih::wrapper::TransactionInputWrapper const &txWrap
              txWrapper.getOptions());
 }
 
-void signTransaction(Transaction &transaction, bytes const& seed)
+// TODO: replace keyFile param with template param once
+// we have key store or other key files implemented
+void signTransaction(Transaction &transaction, PemFileReader const &keyFile)
 {
-    Signer signer(seed);
+    Signer signer(keyFile.getSeed());
     transaction.sign(signer);
 }
 
@@ -58,7 +60,7 @@ void handleCreateSignedTransactionWithPemFile(cxxopts::ParseResult const &result
     PemFileReader pemReader(transactionInputWrapper.getInputFile());
 
     Transaction transaction = internal::createTransaction(transactionInputWrapper, pemReader.getAddress());
-    internal::signTransaction(transaction,pemReader.getSeed());
+    internal::signTransaction(transaction, pemReader);
 
     jsonFile.writeDataToFile(transaction.serialize());
 }
@@ -69,9 +71,10 @@ void handleIssueESDT(cxxopts::ParseResult const &result)
     auto config = data->get_table("Config");
     std::string const networkConfig = *config->get_as<std::string>("NetworkConfig");
 
-    auto userConfig = data->get_table("Testnet");
+    auto userConfig = data->get_table(networkConfig);
     std::string const chainID = *userConfig->get_as<std::string>("ChainID");
     std::string const issueESDTSCAddr = *userConfig->get_as<std::string>("IssueESDTSCAddress");
+    std::string const proxyUrl = *userConfig->get_as<std::string>("ProxyUrl");
 
     std::string const token = result["token-id"].as<std::string>();
     std::string const ticker = result["ticker"].as<std::string>();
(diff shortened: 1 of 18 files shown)


cli/commandhandler.h
```diff
@@ -22,9 +24,11 @@ Transaction createTransaction(ih::wrapper::TransactionInputWrapper const &txWrap
              txWrapper.getOptions());
 }
 
-void signTransaction(Transaction &transaction, bytes const& seed)
+// TODO: replace keyFile param with template param once
```
andreibancioiu, line 27: Well, not necessarily a type parameter, right? Why not receive an interface / abstract class (with several implementations), such as `ISecretKeyProvider`?

cli/commandhandler.h
```diff
@@ -56,11 +60,109 @@ void handleCreateSignedTransactionWithPemFile(cxxopts::ParseResult const &result
...
+    tx.m_gasPrice = gasPrice;
+    tx.m_gasLimit = gasLimit;
+    tx.m_chainID = config.chainID;
+
+    prepareTransactionForESDTTransfer(tx, token, function);
+
+    if (!args.empty())
+    {
+        std::string txData (tx.m_data->begin(), tx.m_data->end());
+        for (auto const &arg : args)
```
andreibancioiu, line 145: Not sure, I think the arguments concatenation / serialization logic is written in other place as well. Perhaps we can extract it and keep it in a single place (in a future PR)?

cli/config/networkconfig.cpp
```diff
@@ -0,0 +1,43 @@
+#include "networkconfig.h"
```
andreibancioiu: 👍 nice for adding this.

However, `NetworkConfig` has a slightly different meaning for the Elrond Gateway, for ERDJS and ERDPY (might generate a bit of confusion). That is, `NetworkConfig` generally means "the structure returned by the API endpoint `network/config`".

Perhaps `AppConfig`?

cli/inputhandler/options_esdt.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    m_optionsIssue.add_options("issue") // esdt issue group
+            ("token", "Token id", cxxopts::value<std::string>())
+            ("ticker", "Token ticker", cxxopts::value<std::string>())
+            ("supply", "Initial supply", cxxopts::value<std::string>())
+            ("dec", "Number of decimals", cxxopts::value<std::string>())
+            ("gas-price", "Gas price", cxxopts::value<uint64_t>())
+            ("pem", "Sender's pem keyfile", cxxopts::value<std::string>())
+            ("nonce", "Sender account's nonce", cxxopts::value<uint64_t>())
+            ("can-freeze", "Can freeze property (default: false)", cxxopts::value<bool>()->default_value("false"))
```
andreibancioiu: `[Can freeze] property`, `[Can wipe] property` (some kind of separators, could be quotes as well). Or simply use a common (and less detailed) description such as `ESDT property`. Or... as it is, but remove "property" :)

cli/inputhandler/options_esdt.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    m_optionsTransfer.add_options("transfer") // esdt transfer group
+            ("nonce", "Sender account's nonce", cxxopts::value<uint64_t>())
+            ("gas-price", "Gas price", cxxopts::value<uint64_t>())
+            ("gas-limit", "Gas limit", cxxopts::value<uint64_t>())
+            ("receiver", "Receiver's address", cxxopts::value<std::string>())
+            ("token", "Token id", cxxopts::value<std::string>())
+            ("value", "ESDT value to transfer", cxxopts::value<std::string>())
+            ("function", "Smart contract function", cxxopts::value<std::string>()->default_value(""))
+            ("args", "Smart contract function args, hex encoded, comma separated", cxxopts::value<std::vector<std::string>>()->default_value(""))
```
andreibancioiu: `... such as: 0xABBA,0xABBA`? If so, perhaps add a dummy example in the description phrase?

andreibancioiu (commented):
Very nice 👍 - as for the previous PR, not sure about the use case - except of serving as an excellent demonstration of the CPP SDK.

cli/commandhandler.h
mariusmihaic: Good point.
I can implement a `ISecretKeyProvider` interface and make `PemFileReader` and `KeyFileReader` derive from it.

cli/commandhandler.h
mariusmihaic: Similar functionality is written in SDK API, using`SCArguments` class

```c++

    SCArguments args;

    BigUInt const bigUInt("10"); //saves internally hex value of decimal 10  = 0a
    Address const address("erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th"); //saves internally hex value of public key = "0139472eff6886771a982f3083da5d421f24c29181e63888228dc81ca60d69e1"
    std::string const str("foo"); //saves internally hex representation of ASCII string "foo" = "666f6f"

    args.add(bigUInt);
    args.add(address);
    args.add(str);
 
    // This will return the concatenated arguments with "@" delimiter
    args.asOnData() == "@0a@0139472eff6886771a982f3083da5d421f24c29181e63888228dc81ca60d69e1@666f6f
```

**However**, for CLI we can **NOT** deduce (in a nice, clean way) the type of the argument. For instance, if CLI --args option looks like this:

```bash
erdcpp esdt transfer --function=someFunction --args=100,erd1j
```

- Was `100` meant to be an ASCII string or a BigUint? 
- Was `erd1j` meant to be a bech32 addres or an ASCII string?

So we just ask the user to fill the **already hex encoded** arguments, so here I add them directly at the end of the data, without the help of `SCArguments`.

cli/config/networkconfig.cpp
mariusmihaic: > +1 nice for adding this.

Thank you ! :)

> However, NetworkConfig has a slightly different meaning for the Elrond Gateway, for ERDJS and ERDPY (might generate a bit of confusion). That is, NetworkConfig generally means "the structure returned by the API endpoint network/config".

Oh, ok, that's my bad, wasn't aware of it

> Perhaps AppConfig?

I think CLIConfig would work in this case, since this config is only meant for CLI

cli/inputhandler/options_esdt.cpp
mariusmihaic: > [Can freeze] property, [Can wipe] property

I applied this suggestion

cli/inputhandler/options_esdt.cpp
mariusmihaic: > ... such as: 0xABBA,0xABBA

Actually just `ABBA`.
I will add a dummy example here.

AdoAdoAdo (dismissed):

andreibancioiu (dismissed):

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #17, opened 2021-07-15
En 10358/read account keyfile

mariusmihaic:
- Added external files ( **NO**  need to review them) for aes-128-ctr decrypt function: 
1. `aes.c`
2. `aes.h`
3. `aes.hpp`
- Added `KeyFileReader` class that can read keyfiles. On its constructor, it expects a `keyfile.json` and its associated `password` to decrypt it and derive the secret key. 
- Added `EncryptedData` and `KdfParams` struct to reflect the keyfile json structure
- Added wrappers for some crypto functions:
1. `getSeed`
2. `scrypt`
3. `hmacsha256`
4. `aes128ctrDecrypt`


include/internal/internal.h
```diff
@@ -1,9 +1,35 @@
...
 #include <vector>
 #include <cstdint>
 
+#define KEY_FILE_VERSION 4U
+#define KEY_FILE_DERIVATION_FUNCTION std::string("scrypt")
+#define KEY_FILE_CIPHER_ALGORITHM std::string("aes-128-ctr")
+
 typedef std::vector<uint8_t> bytes;
 
+struct KdfParams
```
andreibancioiu, line 14: Perhaps also add a reference to this URL?

https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/crypto/derivationParams.ts

include/internal/internal.h
```diff
@@ -1,9 +1,35 @@
...
+{
+    uint32_t dklen = 32;
+    uint32_t n = 4096;
+    uint32_t r = 8;
+    uint32_t p = 1;
+    std::string salt;
+};
+
+struct EncryptedData
```
andreibancioiu, line 23: Perhaps also add a reference to this URL?

https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/crypto/encryptedData.ts

src/utils/errors.h
```diff
@@ -46,18 +30,12 @@ errorMessage const ERROR_MSG_JSON_KEY_NOT_FOUND = "Json does not contain key: ";
 errorMessage const ERROR_MSG_JSON_SET = "Json can not insert key:  ";
 errorMessage const ERROR_MSG_HTTP_REQUEST_FAILED = "Request failed with message: ";
 errorMessage const ERROR_MSG_REASON = "Error reason: ";
+errorMessage const ERROR_MSG_KEY_FILE = "Invalid keyfile.";
+errorMessage const ERROR_MSG_MAC = "MAC mismatch, possibly wrong password.";
+errorMessage const ERROR_MSG_SCRYPTSY = "Scrypt function failed. Could not derive keys, possible cause: operating system refused to allocate the amount of requested memory.";
```
andreibancioiu, line 35: Is this the most probable cause? Why?

tests/test_src/test_filehandler/test_filehandler.cpp
```diff
@@ -94,3 +95,71 @@ TEST_P(PemFileReaderParametrized, getSeed_getPublicKey_getBech32Address)
     EXPECT_EQ(pemPubKey, util::hexToBytes(currParam.publicKey));
     EXPECT_EQ(pemBech32Address, currParam.bech32Address);
 }
+
+class KeyFileReaderConstructorFixture : public ::testing::Test
```
andreibancioiu, line 99: Never thought that CPP unit tests can be written in such a clean and readable way :) :+1:

tests/test_src/test_filehandler/test_filehandler.cpp
```diff
@@ -94,3 +95,71 @@ TEST_P(PemFileReaderParametrized, getSeed_getPublicKey_getBech32Address)
...
+    expectException<std::invalid_argument>("..//..//testData//keyFileInvalidKdf.json", "", ERROR_MSG_KEY_FILE_DERIVATION_FUNCTION);
+}
+
+TEST_F(KeyFileReaderConstructorFixture, invalidContent)
+{
+    expectException<std::invalid_argument>("..//..//testData//keyFileInvalidContent.json", "", ERROR_MSG_KEY_FILE);
+}
+
+// These tests are from :
+// https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/bb926b029150d7c79f2b37308f4334f98a4cabf7/src/testutils/wallets.ts
```
andreibancioiu, line 147: Perhaps use the "main" URL?

https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/testutils/wallets.ts

src/wrappers/cryptosignwrapper.cpp
```diff
@@ -64,6 +79,59 @@ bool verify(std::string const &signature, std::string const &message, bytes cons
     return res == 0;
 }
 
+bytes scrypt(std::string const &password, KdfParams const &kdfParams)
+{
+    unsigned int const keyLength = kdfParams.dklen;
+    unsigned char derivedKey[keyLength];
+
+    auto passw = reinterpret_cast<const unsigned char*>(password.data());
```
andreibancioiu, line 87: Can we extract `reinterpret_cast<const unsigned char*>` to a utility function (or a macro?) called `as_char_ptr` or something like that? (used quite often in this file). Just a suggestion.

src/wrappers/cryptosignwrapper.cpp
```diff
@@ -64,6 +79,59 @@ bool verify(std::string const &signature, std::string const &message, bytes cons
...
+    crypto_auth_hmacsha256_state state;
+
+    crypto_auth_hmacsha256_init(&state, k, key.size());
+    crypto_auth_hmacsha256_update(&state, cipher, cipherText.size());
+    crypto_auth_hmacsha256_final(&state, digest);
+
+    return std::string(digest, digest + HMAC_SHA256_BYTES);
+}
+
+bytes aes128ctrDecrypt(bytes const &key, std::string cipherText, std::string const &iv)
```
andreibancioiu, line 120: Can we perhaps write some "direct" unit tests for this function - and the one above (that is, not only via `deriveSecretKey`)? 

(perhaps not worth the effort, though, it's just parameter forwarding)

src/filehandler/keyfilereader.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+bytes KeyFileReader::getSeed() const
+{
+    return wrapper::crypto::getSeed(m_secretKey);
+}
+
+EncryptedData KeyFileReader::getFileContent() const
+{
+    EncryptedData data;
```
andreibancioiu, line 105: I think we can add the declaration and the return within `try`, as well - just a suggestion.

src/filehandler/keyfilereader.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "errors.h"
+#include "hex.h"
+
+#include <fstream>
+#include <stdexcept>
+
+
+namespace internal
+{
+bytes deriveSecretKey(EncryptedData const &data, std::string const &password)
```
andreibancioiu, line 13: :+1:

andreibancioiu (approved):

AdoAdoAdo (approved):

include/internal/internal.h
mariusmihaic: Added comment

include/internal/internal.h
mariusmihaic: Added comment

src/utils/errors.h
mariusmihaic: This is from libsodium's documentation here: 
https://libsodium.gitbook.io/doc/advanced/scrypt#key-derivation

Which states that: 

> The function returns 0 on success, and -1 if the computation didn't complete, usually because the operating system refused to allocate the amount of requested memory.

tests/test_src/test_filehandler/test_filehandler.cpp
mariusmihaic: Thanks a lot, appreciate it !

tests/test_src/test_filehandler/test_filehandler.cpp
mariusmihaic: Yes, it's better that way, changed it

src/wrappers/cryptosignwrapper.cpp
mariusmihaic: I'm usually not a fan of macros, but here it comes in very handy. 
I've applied your suggestion for this whole file

src/wrappers/cryptosignwrapper.cpp
mariusmihaic: Hmm, would be nice to have some direct tests, however I also don't think it might be worth the effort, since it is indeed just a parameter forwarding

src/filehandler/keyfilereader.cpp
mariusmihaic: Yes, applied your suggestion. Thought it would cause some warnings of type "not all paths return EncryptedData"

src/filehandler/keyfilereader.cpp
mariusmihaic: Thanks !

PR #18, opened 2021-07-19
En 10373/tx hash sign

mariusmihaic:
- Added external files `sha3.h, sha3.c`. **NO** need to review them
- Added `MessageSinger` class which derives from `Signer` class. Its purpose is to **sign/verify** messages with `Elrond Message Prefix`, using sha3 keccak.
- Added functionality to **sign/verify** tx hash instead of serialized transaction, only :
```c++
if (transaction.version == 2 && transaction.options == 1)
```
- Added wrapper for `sha3keccak` function


src/transaction/transaction.cpp
```diff
@@ -25,6 +27,21 @@ void getJsonValueIfNotNull(wrapper::json::OrderedJson const &json, std::string c
...
 }
+
+std::string getTxSerializedHashIfRequired(Transaction const &tx)
+{
+    std::string txSerialized = tx.serialize();
+
+    if( tx.m_options != nullptr &&
+      ( tx.m_version == VERSION_SIGN_TX_HASH) &&
+      (*tx.m_options & OPTIONS_SIGN_TX_HASH_MASK))
+
```
andreibancioiu, line 46: Perhaps extract the conditions to a function called `shouldSignHash` or `isHashSigning` enabled? (ignore if you think it's readable enough, though).

Also, accidental empty line?

src/transaction/transaction.cpp
```diff
@@ -25,6 +27,21 @@ void getJsonValueIfNotNull(wrapper::json::OrderedJson const &json, std::string c
         val = std::make_shared<T>(tmpVal);
     }
 }
+
+std::string getTxSerializedHashIfRequired(Transaction const &tx)
+{
+    std::string txSerialized = tx.serialize();
+
+    if( tx.m_options != nullptr &&
+      ( tx.m_version == VERSION_SIGN_TX_HASH) &&
```
andreibancioiu: `>=` (I think so)?

src/transaction/transaction.cpp
```diff
@@ -75,15 +92,30 @@ Transaction::Transaction() :
...
     m_signature.reset(new std::string(signature));
 }
 
+bool Transaction::verify()
+{
+    if (m_signature == nullptr) throw std::runtime_error(ERROR_MSG_SIGNATURE);
+    if (m_sender == nullptr) throw std::runtime_error(ERROR_MSG_SENDER);
+
+    auto tmp = m_signature;
+    m_signature = nullptr;
```
andreibancioiu: Oh :)

This is because you want to alter the behavior of `serialize()`, right? That fact is not extremely obvious right now, do you think you can perhaps add a comment here to describe the workaround?

src/transaction/transaction.cpp
```diff
@@ -75,15 +92,30 @@ Transaction::Transaction() :
...
+    std::string const txSerialized = internal::getTxSerializedHashIfRequired(*this);
```
andreibancioiu: `getSerialized(asHash: true / false, includeSignature: true / false)` - if this makes sense, perhaps refactor in a future PR?

tests/test_src/test_transaction/test_transaction.cpp
```diff
@@ -5,13 +5,15 @@
...
 #include "transaction/transaction.h"
+#include "transaction/messagesigner.h"
 #include "wrappers/cryptosignwrapper.h"
 
 // Most tests for signing and transaction construction are adapted from one of the following sources:
 // ERD-JS: https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/bb926b029150d7c79f2b37308f4334f98a4cabf7/src/testutils/wallets.ts#L110
 //         https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/walletcore/users.spec.ts#L120
 // ERD-PY: https://github.com/ElrondNetwork/elrond-sdk-erdpy/blob/main/erdpy/tests/test_wallet.py
 // ERD-GO: https://github.com/ElrondNetwork/elrond-go/blob/master/examples/construction_test.go
+//         https://github.com/ElrondNetwork/elrond-go/blob/7d133422048b7e65f0a8730cfdd1fff9d51077e9/examples/messageSign_test.go#L87
```
andreibancioiu, line 16: `main` instead of `7d13...`? Oh, but then the line might change in the future. Can stay as it is.

andreibancioiu (dismissed):
:+1:

external/keccak/sha3.c
```diff
@@ -0,0 +1,191 @@
+// sha3.c
+// 19-Nov-11  Markku-Juhani O. Saarinen <mjos@iki.fi>
+
+// Revised 07-Aug-15 to match with official release of FIPS PUB 202 "SHA3"
+// Revised 03-Sep-15 for portability + OpenSSL - style API
+
+#include "sha3.h"
```
AdoAdoAdo, line 10: please add a comment on top that this file was adapted for keccak, and add what was changed

src/transaction/messagesigner.cpp
```diff
@@ -0,0 +1,40 @@
...
+bool MessageSigner::verify(std::string const &signature, std::string const &message, Address const &address)
+{
+    std::string const hashedMsg = computeHash(message);
+
+    return Signer::verify(signature, hashedMsg, address);
+}
+
+std::string MessageSigner::computeHash(std::string const &message)
+{
+    std::string const erdMsg = ERD_SIGNED_MSG_PREFIX + std::to_string(message.size()) + message;
```
AdoAdoAdo, line 36: can we extract this concat in a different function and call it before calling computeHash?
otherwise the function computeHash is misleading as it is also prefixing the message with the ERD signed message prefix

src/transaction/transaction.cpp
AdoAdoAdo: I would also create a shouldSignHash boolean

src/transaction/transaction.cpp
```diff
@@ -25,6 +27,21 @@ void getJsonValueIfNotNull(wrapper::json::OrderedJson const &json, std::string c
         val = std::make_shared<T>(tmpVal);
     }
 }
+
+std::string getTxSerializedHashIfRequired(Transaction const &tx)
```
AdoAdoAdo: maybe rename to getTxMessageToSign ?

AdoAdoAdo (changes requested):

src/transaction/transaction.cpp
mariusmihaic: I also though in the first place to extract the condition to a boolean function, but dropped the idea meanwhile :D.
I have applied your suggestion and created `shouldSignHash` function.

> Also, accidental empty line?

Yes, forgot about it, fixed !

src/transaction/transaction.cpp
mariusmihaic: Changed it with `>=`

src/transaction/transaction.cpp
mariusmihaic: > This is because you want to alter the behavior of serialize(), right?

Yes, because I want to verify the **signature** of the **message**(the serialized transaction, **without signature**).
Will add a comment here

src/transaction/transaction.cpp
mariusmihaic: - bool parameter `asHash` wouldn't make sense here, since we already verify inside this method if we should return the hash of the transaction or not, now using function `shouldSignHash`
- added `withSignature` bool parameter here, which made code more clear from the caller side. Good suggestion

external/keccak/sha3.c
mariusmihaic: Added comment here !

src/transaction/transaction.cpp
mariusmihaic: Changed its name to `getSerializedTxMsg`. If you consider this name doesn't reflect its content, we could change it

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #19, opened 2021-07-20
En 10438/cli keyfile

mariusmihaic:
**Added**:
- `common.h` file. Contains `checkParam` implementation, which was removed from `KeyFileReader`
- `ISecretKey` pure abstract interface. It serves as an interface for key file readers such as `PemFileReader` and `KeyFileReader`
- `SecretKeyProvider` class. Expects as input (constructor): user CLI options. Holds as internal member a pointer to the interface `ISecretKey`. On its constructor, it checks user inputs and instantiates either a `PemFileReader` or a `KeyFileReader`, otherwise throws error.
- Key + password CLI options
- `cli_utility.h` + `cli_utility.cpp`. It contains two utility functions:
1. `createTransaction` -> Expects as input:  **CLI user inputs** + **CLI config**. Returns  a tx without sender, signature. Tx's chainID si taken from CLI config
2. `signTransaction`->  Expects as input: interface to `ISecretKey` + `Transaction`. Signes tx and adds the sender to it

**Refactored**:
- `IFile` class. Expects as input (constructor) **file path** + optional: **expected extension**. On its constructor, it calls `checkFile` method, which checks if file exists and(if provided) if its extension is as expected.
- `PemFileReader` + `KeyFileReader`. Removed `checkFile` duplicated code, which was doing the same thing for both classes(checking if file exists + file extension). This common implementation is now provided by `IFile`'s constructs.
- `TransactionInputWrapper`. It checks (almost all) tx fields from user inputs. If the inputs are missing, then it returns their default value.
- `commandhandler.h` -> `cli_handler.h` + `cli_handler.cpp`

**Deleted**:
- JsonFile class (`jsonfile.h` + `jsonfile.cpp`)
- PemInputWrapepr class (`pem_input_wrapper.h`)
- `strch.h` file


cli/secretkeyprovider.cpp
```diff
@@ -0,0 +1,40 @@
+#include "secretkeyprovider.h"
+
+#include "utils/errors.h"
+#include "filehandler/pemreader.h"
+#include "filehandler/keyfilereader.h"
+
+SecretKeyProvider::SecretKeyProvider(const cxxopts::ParseResult &input)
```
andreibancioiu, line 7: :+1:

src/filehandler/ifile.cpp
```diff
@@ -1,36 +1,55 @@
...
-{}
+IFile::IFile(std::string path, std::string expectedExtension) :
+        m_filePath(std::move(path)),
+        m_expectedExtension(std::move(expectedExtension))
+{
+    try
+    {
+        IFile::checkFile();
+    }
+    catch (std::exception const &error)
```
andreibancioiu: Perhaps add a comment such as `// We do not handle any exceptions, but re-throw everything instead` - or just drop the `try / catch`?

andreibancioiu (approved):

AdoAdoAdo (dismissed):

andreibancioiu (approved):

AdoAdoAdo (approved):

PR #29, opened 2022-06-28
Token Payment

mariusmihaic:
Added `TokenPayment` class support, which will be used to build `PayLoadData`. Implementation is similar to the one from [erdjs](https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/tokenPayment.ts). However, did not add support for `EGLD` amount yet, on purpose(since it is the native currency, not a token). Since c++ does not have native suport for `BigNumber.Value` as in erdjs, when creating ESDTs and metaESDTs the amount will be a string => extra validity checks had to be added

- Refactored `BigUInt` wrapper to throw exceptions on constructor in case a bad string is provided. Also, negative numbers are no longer valid.
- Added support for `* operator` and `divmod` in `BigUInt`


sstanculeanu (dismissed):

tests/test_src/test_transaction/test_token_payment.cpp
```diff
@@ -0,0 +1,80 @@
...
+    EXPECT_EQ(tokenPayment.toString(), "1");
+    EXPECT_EQ(tokenPayment.toPrettyString(), "1.0 ERDCPP-38f249");
+}
+
+TEST(TokenPayment, invalidAmount)
+{
+    std::string identifier = "USDC-c76f1f";
+    size_t numDecimals = 6;
+
+    EXPECT_THROW(TokenPayment::fungibleFromAmount(identifier, "1.2.3", numDecimals).toPrettyString(), std::invalid_argument);
```
mariusmihaic: Could remove `.toPrettyString()` in this test

src/internal/biguint.cpp
```diff
@@ -35,3 +55,23 @@ const std::string &BigUInt::getValue() const
     return m_value;
 }
 
+BigUInt BigUInt::operator*(const BigUInt &rhs)
```
AdoAdoAdo, line 58: can you add some unit tests?

src/transaction/token_payment.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+    return c == '0';
+}
+
+bool isPoint(char c)
+{
+    return c == '.';
+}
+
+bool hasLeadingZeroes(const std::string &s)
```
AdoAdoAdo, line 19: can you rename to hasInvalidFormat?
would be easier to understand the methods using this.

src/utils/errors.h
```diff
@@ -8,7 +8,7 @@ typedef std::string errorMessage;
 
 errorMessage const ERROR_MSG_EMPTY_VALUE = "Empty value: ";
 errorMessage const ERROR_MSG_NONCE = "Invalid nonce.";
-errorMessage const ERROR_MSG_VALUE = "Invalid value.";
+errorMessage const ERROR_MSG_VALUE = "Invalid value: ";
```
AdoAdoAdo, line 11: should we remove the ':' or '.' from the error messages?

src/internal/biguint.cpp
mariusmihaic: Added support for `==` operator and tests, as well as for` *` and `divmod`

src/transaction/token_payment.cpp
mariusmihaic: The func calling it is called `isValidNumber`.
`isValidNumber` calls `hasLeadingZeroes`, which does not check for valid number format (`isValidNumber` as a whole func does it). `hasLeadingZeroes` simply checks if the number starts with more than one `zero`.
I think the flow as it is now reflects its purpose pretty well.

src/utils/errors.h
mariusmihaic: Yes,  we should definitely change the format as well as the way exceptions are thrown with these message, but this should be done in a separate PR.

AdoAdoAdo (approved):

sstanculeanu (approved):

PR #31, opened 2022-06-30
MultiESDTNFTTransfer payload builder

mariusmihaic:
- Added support for `MultiESDTNFTTransfer` payload builder


tests/test_src/test_transaction/test_payload_builder.cpp
```diff
@@ -36,3 +36,26 @@ TEST(ESDTNFTTransferPayloadBuilder, build)
             .build();
     EXPECT_EQ(payload, "ESDTNFTTransfer@4c4b4d45582d616162393130@3cc98c@0b7de7c1695eab800000@000000000000000005005f964241c0607e1c02777654f30072f03dd0abcd5483");
 }
+
+TEST(MultiESDTNFTTransferPayloadBuilder, build)
```
dragos-rebegea, line 40: maybe add some tests also for SFT & metaEsdt

mariusmihaic: This test already checks for metaESDTs(see `LKMEX-aab910` and `LKFARM-9d1ea8`).

However, IMHO it would not bring any more benefits to add extra tests for other tokens. Reason being that those test would only test sc args enconding and tokens encoding is already being tested through other tests. This test should only/mostly test the correct parameter assembly order.

dragos-rebegea: agree!

dragos-rebegea (approved):

sstanculeanu (approved):

PR #34, opened 2022-07-04
Transaction factory egld + esdt transfer

mariusmihaic:
Added `TransactionFactory` which creates transaction builders for each type of transactions, in form in abstract classes, which cannot be used/instantianted on their own. (in this PR only `EGLDTransfer` and `ESDTTransfer` are implemented, more will follow in other PRs)

- Added abstract builder classes `ITransactionBuilder` (for egld type transfers) and `ITokenTransactionBuilder` (for tokens type transfers). Both builders have different options to be added to transactions, e.g.: `withVersion`, `withContractCall`, etc. Each transaction can be `build` without signatures, or `buildSigned` with a private key/ wallet.
- Added convenient struct paceholder for tx builders input: `TransactionBuilderInput`
- Github actions (build + tests) are triggered for each opened PR
- Fixed a bug in `ContractCall::asOnData()` to also add `@` in front of the encoded string if not empty


include/transaction/transaction_builders.h
```diff
@@ -0,0 +1,44 @@
+#ifndef ERD_TRANSACTION_BUILDER_ESDT_H
+#define ERD_TRANSACTION_BUILDER_ESDT_H
```
andreibancioiu, line 2: This does not correspond to the name of the file.

tests/test_src/test_transaction/test_transaction_factory.cpp
```diff
@@ -0,0 +1,105 @@
+#include "gtest/gtest.h"
+#include "filehandler/pemreader.h"
+#include "transaction/transaction_factory.h"
+#include "test_common.h"
+
+TEST(TransactionFactory, createEGLDTransfer)
```
andreibancioiu, line 6: :+1:

src/smartcontracts/contract_call.cpp
```diff
@@ -18,5 +18,6 @@ void ContractCall::setArgs(SCArguments args)
 
 std::string ContractCall::asOnData() const
 {
-    return util::stringToHex(m_function) + m_args.asOnData();
+    return (m_function.empty()) ?
```
andreibancioiu, line 21: A bit hack-ish (not obvious that `m_args.asOnData()` has a `@` prefix, as well), but can stay as it is for the moment.

src/transaction/transaction_factory.cpp
```diff
@@ -0,0 +1,47 @@
+#include <utility>
+
```
andreibancioiu, line 2: At this point, I'm not sure whether defining only builders (no factory) would have been friendly-enough for developers.

andreibancioiu (approved):

dragos-rebegea (approved):

include/transaction/transaction_builders.h
mariusmihaic: Will change it in a future PR, thanks! :+1:

tests/test_src/test_transaction/test_transaction_factory.cpp
mariusmihaic: :pray:

src/smartcontracts/contract_call.cpp
mariusmihaic: I actually think `asOnData` should return prefixed data with `@`, since that is the way it always looks in data field. 
E.g.:
1. @04@f4f4f4@3f -> all args have `@`
2. 04@f4f4f4@3f -> only last 2 args have `@`, does not seem consistent

src/transaction/transaction_factory.cpp
mariusmihaic: I'm not that sure either, but that seemed the best solution atm

PR #38, opened 2022-07-07
Transaction builder ESDT issue

mariusmihaic:
- Added `gasLimit` field in `TxBuilderInput`, which is by default = `DEFAULT_GAS_LIMIT`. If gas limit input from factory !=  `DEFAULT_GAS_LIMIT`, tx builder will use that one, otherwise, it will compute it using the `gasEstimator`
- Added esdt issue payload builder
- Added tx builder


src/transaction/payload_builder.cpp
```diff
@@ -1,9 +1,33 @@
...
+
+std::string ESDTPropertyField(std::string const &property, bool const &val)
+{
+    std::string const value = val ? "true" : "false";
+    return "@" + util::stringToHex(property) + "@" + util::stringToHex(value);
+}
+
+std::string ESDTPropertiesAsOnData(ESDTProperties const &esdtProperties)
+{
+    return ESDTPropertyField("canFreeze", esdtProperties.canFreeze) +
```
dragos-rebegea, line 19: add canTransferNFTCreateRole

src/transaction/payload_builder.cpp
```diff
@@ -103,3 +127,47 @@ std::string MultiESDTNFTTransferPayloadBuilder::build() const
...
+std::string ESDTIssuePayloadBuilder::build() const
+{
+    SCArguments args;
+    args.add(m_token);
+    args.add(m_ticker);
+    args.add(m_initialSupply);
+    args.add(BigUInt(m_numOfDecimals));
+
+    std::string data = ESDT_ISSUANCE_PREFIX + args.asOnData();
+    data += (m_esdtProperties != ESDT_ISSUANCE_DEFAULT_PROPERTIES) ? ESDTPropertiesAsOnData(m_esdtProperties) : "";
```
dragos-rebegea, line 171:
```suggestion
    if (m_esdtProperties != ESDT_ISSUANCE_DEFAULT_PROPERTIES) {
        data += ESDTPropertiesAsOnData(m_esdtProperties)
    }
```

dragos-rebegea (changes requested):

src/transaction/payload_builder.cpp
mariusmihaic: Right, was not aware of it. Thanks, good suggestion :+1:

src/transaction/payload_builder.cpp
mariusmihaic: Matter of preference, I would keep it in 1 line as it is

dragos-rebegea: :+1:

dragos-rebegea (approved):

include/transaction/payload_builder.h
```diff
@@ -61,4 +62,26 @@ class MultiESDTNFTTransferPayloadBuilder
     ContractCall m_contractCall;
 };
 
+class ESDTIssuePayloadBuilder
```
andreibancioiu, line 65: :rocket:

src/transaction/transaction_builders.cpp
```diff
@@ -12,14 +12,17 @@ std::shared_ptr<bytes> stringToBytesPtr(std::string const &in)
...
         ITransactionBuilder(),
         m_txInput(std::move(txInput))
 {}
 
 Transaction TransactionEGLDTransferBuilder::build()
 {
-    uint64_t gasLimit = m_txInput.gasEstimator.forEGLDTransfer(m_txInput.data.size());
+    uint64_t inputGasLimit = m_txInput.gasLimit;
+    uint64_t estimatedGasLimit = m_txInput.gasEstimator.forEGLDTransfer(m_txInput.data.size());
+    uint64_t gasLimit = (inputGasLimit == DEFAULT_GAS_LIMIT) ? estimatedGasLimit : inputGasLimit;
```
andreibancioiu, line 25: Somehow, if the user provides a gas limit equal to DEFAULT_GAS_LIMIT, that's ignored. However, the result is still good :)

So, tricky, but correct - can you double check, though?

andreibancioiu (approved):
:+1:

src/transaction/transaction_builders.cpp
mariusmihaic: User cannot provide gasLimit, only gasPrice. GasLimit is always set by the factory in builders

PR #39, opened 2022-07-11
Replace prepareTransactionFor... functions with factory

mariusmihaic:
Completely removed `prepareTransactionForESDTTransfer`/`prepareTransactionForESDTIssuance` along with their associated tests, since we have `TransactionFactory`, which builds transactions instead.

- Removed `IssueESDTSCAddress` from `CliConfig`, since the address is a system smart contract, so it will have the same value for each network (redundancy)
- Refactored integration tests from `test_proxyprovider.cpp` to use `TransactionFactory` instead of deleted obsolete functions
- Tested cli refactor on testnet: https://testnet-explorer.elrond.com/accounts/erd1sjsk3n2d0krq3pyxxtgf0q7j3t56sgusqaujj4n82l39t9h7jers6gslr4. More PRs will follow to make an integration test for CLI.


cli/cli_handler.cpp
```diff
@@ -44,7 +44,7 @@ void handleIssueESDT(cxxopts::ParseResult const &result)
     auto const token = result["token"].as<std::string>();
     auto const ticker = result["ticker"].as<std::string>();
     auto const supply = result["supply"].as<std::string>();
-    auto const decimals = result["dec"].as<std::string>();
+    auto const decimals = result["dec"].as<uint32_t>();
```
dragos-rebegea, line 47: we can use also uint8_t

cli/inputhandler/options_esdt.cpp
```diff
@@ -35,7 +35,7 @@ void OptionsESDT::initOptionsIssue()
             ("token", "Token id", cxxopts::value<std::string>())
             ("ticker", "Token ticker", cxxopts::value<std::string>())
             ("supply", "Initial supply", cxxopts::value<std::string>())
-            ("dec", "Number of decimals", cxxopts::value<std::string>())
+            ("dec", "Number of decimals", cxxopts::value<uint32_t>())
```
dragos-rebegea, line 38: we can use also uint8_t

dragos-rebegea (approved):
We decided to use uint32_t for decimals (even if uin8_t would also fit for values <= 18) as it is on the elrond-go

andreibancioiu (approved):

PR #40, opened 2022-07-11
Replace string with BigUInt

mariusmihaic:
Replaced balance/value related fields with `BigUInt`, instead of `string` in:
- `ESDT` balance/(s)
- Transaction `value`
- Account `balance`

Additionally:
- Removed `TransactionHash`, which only contained `string hash`
- Added support for `/` operator in `BigUInt`

### DIFF
--- a/cli/cli_handler.cpp
+++ b/cli/cli_handler.cpp
@@ -83,7 +83,7 @@ void handleIssueESDT(cxxopts::ParseResult const &result)
             esdtProperties
     )->buildSigned(keyFile->getSeed());
 
-    auto const txHash = proxy.send(tx).hash;
+    auto const txHash = proxy.send(tx);
     std::cerr << "Transaction hash: " << txHash << "\n";
 }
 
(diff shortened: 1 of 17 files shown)


cli/cli_handler.cpp
```diff
@@ -83,7 +83,7 @@ void handleIssueESDT(cxxopts::ParseResult const &result)
             esdtProperties
     )->buildSigned(keyFile->getSeed());
 
-    auto const txHash = proxy.send(tx).hash;
+    auto const txHash = proxy.send(tx);
```
andreibancioiu, line 86: For future - there's also an endpoint that allows one to send multiple transactions: `/transaction/send-multiple` (erdpy uses it, erdjs does not - yet).

include/internal/biguint.h
```diff
@@ -10,7 +10,9 @@ class BigUInt
 
     explicit BigUInt(std::string value);
 
-    BigUInt operator*(BigUInt const &rhs);
+    BigUInt operator*(BigUInt const &rhs) const;
```
andreibancioiu, line 13: They do not seem to be used now (sorry if mistaken). But indeed, can be useful to compute, for example, fees (in the future).

andreibancioiu (approved):

cli/cli_handler.cpp
mariusmihaic: Good point, could add a task for this, thanks! :+1:

include/internal/biguint.h
mariusmihaic: Yes, not used in project for now, but useful for those who wish to use it.

dragos-rebegea (approved):

PR #41, opened 2022-07-12
CLI - set network

mariusmihaic:
Added support for CLI to set the current network config, which will be used to interact with it.

### DIFF
--- a/cli/cli_handler.cpp
+++ b/cli/cli_handler.cpp
@@ -134,28 +134,33 @@ void handleSetNetworkConfig(cxxopts::ParseResult const &result)
                    [](unsigned char c)
                    { return std::tolower(c); });
 
-    auto const config = CLIConfig();
+    auto const cliConfig = CLIConfig();
     network[0] = char(toupper(network[0]));
     if (network == NETWORK_MAINNET)
     {
-        config.setNetwork(Mainnet);
+        cliConfig.setNetwork(Mainnet);
     }
     else if (network == NETWORK_DEVNET)
     {
-        config.setNetwork(Devnet);
+        cliConfig.setNetwork(Devnet);
     }
     else if (network == NETWORK_TESTNET)
     {
-        config.setNetwork(Testnet);
+        cliConfig.setNetwork(Testnet);
     }
     else if (network == NETWORK_LOCAL)
     {
-        config.setNetwork(Local);
+        cliConfig.setNetwork(Local);
     }
     else
     {
-        std::cerr << "Invalid network\n";
+        throw std::invalid_argument("Invalid network: " + network);
     }
+
+    auto cfg = cliConfig.config();
+    std::cerr << "Config changed to: " << network << "\n";
+    std::cerr << "ChainID = " << cfg.chainID << "\n";
+    std::cerr << "ProxyUrl = " << cfg.proxyUrl << "\n";
 }
 
 void handleRequest(ih::ArgParsedResult const &parsedResult)
--- a/cli/cli_handler.cpp
+++ b/cli/cli_handler.cpp
@@ -127,6 +127,37 @@ void handleTransferESDT(cxxopts::ParseResult const &result)
     std::cerr << "Transaction hash: " << txHash << "\n";
 }
 
+void handleSetNetworkConfig(cxxopts::ParseResult const &result)
+{
+    std::string network = result["config"].as<std::string>();
+    std::transform(network.begin(), network.end(), network.begin(),
+                   [](unsigned char c)
+                   { return std::tolower(c); });
+
+    auto const config = CLIConfig();
+    network[0] = char(toupper(network[0]));
+    if (network == NETWORK_MAINNET)
+    {
+        config.setNetwork(Mainnet);
+    }
+    else if (network == NETWORK_DEVNET)
+    {
+        config.setNetwork(Devnet);
+    }
+    else if (network == NETWORK_TESTNET)
+    {
+        config.setNetwork(Testnet);
+    }
+    else if (network == NETWORK_LOCAL)
+    {
+        config.setNetwork(Local);
+    }
+    else
+    {
+        std::cerr << "Invalid network\n";
+    }
+}
+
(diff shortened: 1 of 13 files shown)


cli/config/cliconfig.cpp
```diff
@@ -1,15 +1,29 @@
...
 
 #define BEGIN  "\x1B[33m"
 #define END  "\x1B[0m"
 #define YELLOW(x) BEGIN x END
 
-CLIConfig::CLIConfig(std::string tomlConfigPath):
-    m_tomlPath(std::move(tomlConfigPath)) {}
+namespace
+{
+inline std::string fileDoesNotContain(std::string const& filePath)
```
andreibancioiu: `formatMessageFileDoesNotContain`?

andreibancioiu: Or, perhaps, even better, define custom errors?

cli/config/cliconfig.cpp
```diff
@@ -38,3 +60,41 @@ Config CLIConfig::config() const
...
+{
+    std::ifstream inFile(m_tomlPath);
+    std::stringstream strStream;
+
+    strStream << inFile.rdbuf();
+    std::string currLine;
+    bool found = false;
+    while (std::getline(strStream, currLine))
+    {
+        if (currLine.find("CLIConfig") != std::string::npos)
```
andreibancioiu, line 73: A bit tricky. Cannot use the toml parser (&serializer) here?

cli/config/cliconfig.cpp
```diff
@@ -1,15 +1,29 @@
...
+namespace
+{
+inline std::string fileDoesNotContain(std::string const& filePath)
+{
+    return "File " + filePath + " does not contain ";
+}
+}
+
+CLIConfig::CLIConfig(std::string const &tomlConfigPath) :
+        m_tomlPath(util::getCanonicalRootPath(tomlConfigPath)),
```
andreibancioiu, line 21: Why not look in the current working directory?

src/utils/common.h
```diff
@@ -24,6 +24,19 @@ inline void checkParam<std::string>(std::string const &param, std::string const
...
     }
 }
+
+inline std::string getCanonicalRootPath(std::string const &path)
+{
+    // Get absolute path to executable
+    std::string canonicalPath = std::string(canonicalize_file_name("/proc/self/exe"));
+
+    // Remove everything in path until elrond-sdk-erdcpp directory and concatenate it with the path
+    // Use rfind because github action runs into elrond-sdk-erdcpp/elrond-sdk-erdcpp folder
```
andreibancioiu, line 34: Altering the github workflow is preferred to customizing the code on this purpose (no action necessary, though).

cli/cli_handler.cpp
```diff
@@ -127,6 +127,42 @@ void handleTransferESDT(cxxopts::ParseResult const &result)
     std::cerr << "Transaction hash: " << txHash << "\n";
 }
 
+void handleSetNetworkConfig(cxxopts::ParseResult const &result)
```
andreibancioiu, line 130: Do we have a README for the cli interaction, as well (can be added in a future PR)?

I'm not sure whether the CLI is for testing the SDK or it's destined to become a swiss-knife for interacting with the network (similar to erdpy).

andreibancioiu (commented):
Just a few questions / remarks.

cli/config/cliconfig.cpp
mariusmihaic: Renamed to  `formatMessageFileDoesNotContain`.

>Or, perhaps, even better, define custom errors?
Not really doable, it's pretty specific for this file only.

cli/config/cliconfig.cpp
mariusmihaic: Unfortunately, not. The library has pretty limited usage, so it does not support changing files... Decided to go this way instead of changing the library.
If, in the future we'll need more changes, I will go with another library

cli/config/cliconfig.cpp
mariusmihaic: Could not, unfortunately. That's because the path is always relative to the executable calling it. So, if your run this from a test file, it would fail. It would also fail from github actions for example

src/utils/common.h
mariusmihaic: This works either way, `rfind` just makes sure that if multiple `elrond-sdk-erdcpp` folders exist, it will choose the utter most right one.
Agree regarding github customizing, but couldn't find how to change the root path. Do you know what's the trick?

cli/cli_handler.cpp
mariusmihaic: > Do we have a README for the cli interaction, as well (can be added in a future PR)?

Yes, but only shows what the help command do: https://github.com/ElrondNetwork/elrond-sdk-erdcpp/blob/development/examples/examples.md#22-examples-cli. I have a separate task to update it, since new features came along

> I'm not sure whether the CLI is for testing the SDK or it's destined to become a swiss-knife for interacting with the network (similar to erdpy).

No, it's not for testing, just similar to erdpy, but very limited usage.

cli/config/cliconfig.cpp
andreibancioiu: In order to change the file = deserialize, change in memory, serialize, then save to file (without special support from the library).

src/utils/common.h
andreibancioiu: Building, then copying the binary to a custom location (and changing the directory) is supported by Github workflows. Even though changing the path of the repository is a bit trickier (I think you need to configure https://github.com/actions/checkout).

andreibancioiu (approved):

cli/config/cliconfig.cpp
mariusmihaic: Yes, but there is no native, nor easy support for serialization/deserialization in c++. It would've required too much effort instead of simply doing like this for this sole purpose

dragos-rebegea (approved):

PR #42, opened 2022-07-20
Add script to install library

mariusmihaic:
Added simple script to install all necessary dependencies for erdcpp.
After building the solution, it will copy the include headers + generated static libraries in `/usr` folder

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,21 @@
+CURRENT_DIR=$(pwd)
+INCLUDE_PATH=/usr/include/
+LIB_PATH=/usr/lib
+
+echo "Installing libsodium library..."
+sudo apt install libsodium-dev
+
+echo "Updating submodules..."
+git submodule update --init --recursive
+
+echo "Cmake"
+cmake .
+
+echo "Building solution..."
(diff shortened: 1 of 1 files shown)


sstanculeanu (approved):

install.sh
```diff
@@ -0,0 +1,21 @@
...
+cmake .
+
+echo "Building solution..."
+cmake --build .
+
+echo "Installing sdk..."
+sudo cp -r "$CURRENT_DIR/include/" "$INCLUDE_PATH/erdcpp"
+sudo cp "$CURRENT_DIR/src/libsrc.so" "$LIB_PATH"
+
+echo "Finished successfully !"
```
bogdan-rosianu, line 21: might add a more detailed message, like `"Finished successfully ! You can now import erdcpp by using '#include erdcpp.h'"` or something similar

bogdan-rosianu (approved):

install.sh
mariusmihaic: Right, this is the next task in updating the `README`. Will add this comments and how to import the library using cmake.

PR #44, opened 2022-07-21
Update external libs + make compatible with newer versions

mariusmihaic:
- Update external libs (everything under `external` directory should not be reviewed)
- Fixed several compile warnings/errors for newer/older versions, e.g.: include extra `toml.h` file to compile `cpptoml.h` with newer versions, add `stdexcept.h`, add `inline` funcs
- Updated and fixed  `install.sh` script and yaml files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.11)
 set(CMAKE_BUILD_TYPE Debug)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
+set(CMAKE_CXX_FLAGS "-fPIC")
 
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.11)
 set(CMAKE_BUILD_TYPE Debug)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
+set(CMAKE_CXX_FLAGS "-fPIC")
```
AdoAdoAdo, line 6: :+1:

raduchis (approved):
install.sh works and new projects using erdcpp do compile

AdoAdoAdo (approved):

